PR #4, opened 2020-08-27
new pulpcore 3.6 deps

evgeni:


evgeni:
I need to fix some deps

ehelms (approved):
If repoclosure is expected

evgeni:
@ehelms a few of them were not, but the last push should fix all of them, we'll see :)

evgeni:
Now the repoclosure is "expected" (as in: why do I still have those packages here)

PR #14, opened 2020-09-03
Bump python-pulp-deb to 2.6.1

evgeni:


packages/python-pulp-deb/python-pulp-deb.spec
```diff
@@ -1,36 +1,38 @@
...
-Version:        2.4.0
-Release:        0.1.b1%{?dist}
+Version:        2.6.1
+Release:        1%{?dist}
 Summary:        pulp-deb plugin for the Pulp Project
 
 License:        GPLv2+
-URL:            https://pulpproject.org/#deb
-Source0:        https://files.pythonhosted.org/packages/source/p/%{pypi_name}/%{pypi_name}-%{version}b1.tar.gz
+URL:            https://pulpproject.org
```
evgeni, line 10: @quba42 if you want to keep that `#deb` anchor you had before, please set the anchored URL in setup.py, then the script will automatically update it :)

quba42: From opening it just now, it does not look like the `#deb` anchor still leads anywhere useful, so I guess it is fine to drop it?

quba42 (approved):
Looks sane to me, but I am no packaging expert :wink:.

packages/python-pulp-deb/python-pulp-deb.spec
evgeni: Probably, yeah.

What I really wanted to say: if you ever have a more useful page than the main homepage of pulp, just drop it into setup.py and the rest will happen automatically :)

quba42: I will try to keep it in mind. :wink:

packages/python-pulp-deb/python-pulp-deb.spec
```diff
@@ -1,36 +1,38 @@
...
+Source0:        https://files.pythonhosted.org/packages/source/p/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
```
ekohl, line 11: @evgeni Any idea why pyp2rpm doesn't use `%{pypi_source}`? It'd be much nicer.

ekohl (approved):

packages/python-pulp-deb/python-pulp-deb.spec
evgeni: I think I disabled that in our template as it didn't work for the modules we rename. Or something like that. we can revisit that

PR #16, opened 2020-09-04
add package automation scripts and templates

evgeni:


automation/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+podman build --tag tfm-pulpcore-builder .
+```
+
+### Run
+
+1. create an empty directory to host the resulting packages
+2. pass the directory as a volume for `/app/packaging` to the container:
+```
+podman run --rm --volume $(pwd)/packaging/:/app/packaging/:Z  --env-file packaging.env tfm-pulpcore-builder
```
ekohl, line 37: I get SELinux denials when I share volumes. Is there some boolean I'm missing?

automation/_generate_deps.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+PULPCORE_PACKAGES="pulpcore==3.6.2 pulp-file==1.2.0 pulp-container==2.0.0 pulp-deb==2.6.1 pulp-rpm==3.6.1 pulp-certguard==1.0.2 galaxy-importer==0.2.8 pulp-ansible==0.2.0"
```
ekohl: Is it time for a requirements.txt file?

automation/README.md
evgeni: Can you post them? The above works on my F32 with Selinux in enforcing when `$(pwd)/packaging/` is a newly created dir in my home

automation/_generate_deps.sh
evgeni: modern shit, but yeah I guess

automation/README.md
ekohl: Ah, I didn't see the `:Z`. Looks like that changes the file context:
```console
$ mkdir packaging
$ ls -lZd packaging/
drwxrwxr-x. 2 ekohl ekohl unconfined_u:object_r:user_home_t:s0 4096 Sep  4 10:55 packaging/
$ podman run --rm --volume $(pwd)/packaging/:/app/packaging/:Z -it centos:7
[root@b240c45dee80 /]# exit
$ ls -lZd packaging/
drwxrwxr-x. 2 ekohl ekohl system_u:object_r:container_file_t:s0:c123,c1003 4096 Sep  4 10:55 packaging/
```

automation/_generate_deps.sh
```diff
@@ -0,0 +1,39 @@
...
+
+  git config --local user.email "${EMAIL:-root@localhost}"
+  git config --local user.name "${NAME:-root}"
+  export RPM_PACKAGER="${NAME}"
+
+  while read line; do
+    pkg=${line%==*}
+    version=${line#*==}
+    REWRITE_ON_SAME_VERSION=false $FOREMAN_PACKAGING/add_pypi_package.sh ${pkg} ${version} $PULPCORE_TAG '/' $PULPCORE_PACKAGING/pyp2rpm/pulpcore.spec
+    sleep 5
```
ekohl, line 38: Is this to prevent pypi from ratelimiting you? Can use a comment :)

pyp2rpm/pulpcore.spec
```diff
@@ -0,0 +1,105 @@
+{{ data.credit_line }}
```
ekohl, line 1: Should this use a `.j2` extension to indicate it's a template? Is that a common convention?

automation/README.md
evgeni: Yes, the :Z enforces it to be the right context for running the container.

pyp2rpm/pulpcore.spec
evgeni: We had that discussion when we added those templates in `foreman-packaging` initially: https://github.com/theforeman/foreman-packaging/pull/4318#discussion_r346842455

TL;DR: pyp2rpm would break :(

automation/README.md
ekohl: I learned something new today :)

automation/_generate_deps.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
ekohl, line 1:
```suggestion
#!/bin/bash -e
```

automation/README.md
evgeni: The best part of it: on systems without SELinux passing :Z (or :z) to --volume breaks Docker/podman, so you need to do funny hacks if you want scripts to work on both worlds:

https://github.com/Katello/katello-client-bootstrap/blob/master/Makefile#L5

ehelms: Would a little wrapper script that creates a default directory for me and just runs this command be asking too much?

evgeni: Absolutely!

automation/_generate_deps.sh
evgeni: I prefer `set -e` instead, done

automation/update_packaging.sh
```diff
@@ -0,0 +1,16 @@
...
+
+mkdir -p ${RESULT_DIR}
+
+if [[ -f ${ENV_FILE} ]]; then
+  ENV_FILE_ARG="--env-file ${ENV_FILE}"
+else
+  ENV_FILE_ARG=""
+fi
+
+podman run --rm --volume ${RESULT_DIR}:/app/packaging/:Z  ${ENV_FILE_ARG} tfm-pulpcore-builder
```
ekohl: Now that there's a script, do you want to apply the `:Z` conditional based on SELinux as well?

automation/README.md
ehelms, line 37: Ought this mention the script now first?

ehelms (approved):

PR #21, opened 2020-09-09
patch django for FIPS

mdellweg:
Not yet confirmed, this is the (only) patch we need.

But i don't quite remember if specifying the patch as an additional source in the spec is sufficient to get it applied.

### DIFF
--- a/packages/python-django/python-django.spec
+++ b/packages/python-django/python-django.spec
@@ -4,12 +4,13 @@
 
 Name:           python-%{srcname}
 Version:        2.2.16
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        A high-level Python Web framework that encourages rapid development and clean, pragmatic design
 
 License:        BSD
 URL:            https://www.djangoproject.com/
 Source0:        https://files.pythonhosted.org/packages/source/D/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+Patch0:         0001-FIPS-Mark-use-of-MD5-not-security-relevant.patch
 BuildArch:      noarch
 
 BuildRequires:  python3-devel
@@ -31,7 +32,7 @@ Requires:       python3-sqlparse >= 0.2.2
 %{summary}
 
 %prep
-%autosetup -n %{pypi_name}-%{version}
+%autosetup -p1 -n %{pypi_name}-%{version}
 # Remove bundled egg-info
 rm -rf %{pypi_name}.egg-info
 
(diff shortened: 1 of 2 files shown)


bmbouter (approved):
This code looks good to me.

packages/python-django/python-django.spec
```diff
@@ -10,6 +10,7 @@ Summary:        A high-level Python Web framework that encourages rapid developm
 License:        BSD
 URL:            https://www.djangoproject.com/
 Source0:        https://files.pythonhosted.org/packages/source/D/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+Source1:        0001-FIPS-Mark-use-of-MD5-not-security-relevant.patch
```
evgeni:
```suggestion
Patch0:        0001-FIPS-Mark-use-of-MD5-not-security-relevant.patch
```

evgeni: also needs to change

```
%autosetup -n %{pypi_name}-%{version}
```

to

```
%autosetup -n %{pypi_name}-%{version} -p1
```

(I guess -p1, based on the diff)

evgeni:
@mdellweg yeah, now add `-p1` to the `%autosetup` call and it will build again

mdellweg:
Thanks @evgeni 
Only when a test was read and turns back green, you know you fixed something. ;)

evgeni:
Okay. This looks correct now, just needs a release bump and a changelog entry.

This should have tripped the CI, wonder why it didn't...

mdellweg:
Right, where was my head today?

evgeni:
My best guess would be Munich, but what do I know?! ;)

packages/python-django/python-django.spec
```diff
@@ -4,12 +4,13 @@
...
 Name:           python-%{srcname}
 Version:        2.2.16
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        A high-level Python Web framework that encourages rapid development and clean, pragmatic design
 
 License:        BSD
 URL:            https://www.djangoproject.com/
 Source0:        https://files.pythonhosted.org/packages/source/D/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+Patch0:        0001-FIPS-Mark-use-of-MD5-not-security-relevant.patch
```
evgeni:
```suggestion
Patch0:         0001-FIPS-Mark-use-of-MD5-not-security-relevant.patch
```

picky ;-)

evgeni:
packaging wise this LGTM

packages/python-django/python-django.spec
mdellweg: :roll_eyes:

mdellweg:
I am quite confident now, this _is_ the only thing we need to patch.
Thenceforth I am putting it out of draft mode.

evgeni (approved):

PR #26, opened 2020-09-14
Deps 20200914

evgeni:


evgeni:
fao89 reported on IRC that drf-spectacular might be breaking things, so please don't merge.

evgeni:
obsolete

PR #33, opened 2020-09-29
pulpcore 3.7 and plugins

evgeni:


packages/python-pulp-ansible/python-pulp-ansible.spec
```diff
@@ -50,6 +50,12 @@ rm -rf %{pypi_name}.egg-info
 %{python3_sitelib}/pulp_ansible-%{version}-py%{python3_version}.egg-info
 
 %changelog
+* Thu Oct 01 2020 Evgeni Golov - 1:0.4.1-1
+- Release python-pulp-ansible 0.4.1
+
+* Mon Sep 28 2020 Evgeni Golov 1:0.4.0-1
+- Update to 0.4.0
+
```
ekohl, line 58: I wonder how useful this entry is

ekohl (approved):
Small nit, otherwise it looks ok

packages/python-pulp-ansible/python-pulp-ansible.spec
evgeni: It's not, just tools doing tools jobs ;)

PR #37, opened 2020-10-06
add pulpcore-wrappers package

evgeni:


evgeni:
@ekohl @mikedep333 so here is a draft of the "packaged wrappers" we talked about earlier.

I mostly wonder about one thing (well, two, dependant):
* should this be an own package or live in `pulpcore-selinux`?
* if it's an own package, should it be `pulpcore` instead of `pulpcore-wrappers` and contain more *common* things for deployments, like user-creation etc.

I am slightly leaning towards "put that into `pulpcore-selinux` and forget about it", but at the same time wanted y'all input.

packages/pulpcore-wrappers/pulpcore-wrappers.spec
```diff
@@ -0,0 +1,42 @@
...
+%description
+pulpcore SELinux wrappers
+
+%prep
+# nothing
+
+%build
+for wrapper in %{wrappers}
+do
+  printf '#!/bin/bash\nexec %s "$@"\n"' ${wrapper} > ${wrapper}
```
ekohl: The number of `"` is 3. Is that correct?

evgeni: sure ;)

ekohl:
Making it part of selinux feels a bit odd to me. You could have installation instructions that leave out pulpcore-selinux if SELinux is off. In that case, you need to modify your systemd templates to use `/usr/bin` paths. However, if you unconditionally install the wrappers there's no need to care about templating the systemd unit files. This could bring us closer to the point where we can share the unit files between both installers.

I'm leaning to pulpcore itself, possibly as a subpackage.

evgeni:
There is no `pulpcore` package today. Or do you mean `python-pulpcore`?

mikedep333:
Honestly, I assumed this would be part of the pulpcore (python-pulpcore) package.

I don't think it makes a big difference. We could always make it part of it later.

The pulpcore-selinux package should depend on it though.

packages/pulpcore-wrappers/empty-file
```diff
@@ -0,0 +1 @@
+# really
```
ehelms, line 1: What's this fella for?

ekohl: I saw @evgeni complain about a missing Source0. I guess this is the hack to allow it to build. If it's done as part of python3-pulpcore, the problem goes away because that has a source.

It would also solve https://pulp.plan.io/issues/7663 if the installers are updated to use the libexec path.

ekohl:
This should be updated to also restore `pulpcore-wrappers` here (if we do indeed go with `pulpcore-wrappers`):
https://github.com/theforeman/pulpcore-packaging/blob/faabec8978c1cd9166da3fb715b7a97b3080c19c/packages/pulpcore-selinux/pulpcore-selinux.spec#L60

If they're included in `python3-pulpcore`, there's no change needed in packaging, just the installers. For that, it would be very convenient if the change was backported to 3.6.

mikedep333:
@ekohl  I just released pulpcore-selinux 1.1.2 .
https://github.com/pulp/pulpcore-selinux/releases/tag/1.1.2

You'll need it for the gunicorn wrapper to get the correct label.

evgeni:
#38 is better and merged, closing

PR #38, opened 2020-10-07
Add libexec wrappers to python-pulpcore

ekohl:
Using these makes it easier for an application to run in SELinux confined mode. pulpcore-selinux only relabels files owned by python3-pulpcore. That means /usr/bin/{gunicorn,rq} are not relabeled if they're already installed. The result is that Pulp runs unconfined.

By having these wrappers, they're guaranteed to contain the correct SELinux context. They are also a good place for additional patches if they are ever needed.

It is an alternative to https://github.com/theforeman/pulpcore-packaging/pull/37. There is the consideration that now the file can no longer be generated by pip2rpm, which pulpcore-wrapper has no problem with.

It also bumps pulpcore-selinux to 1.1.2 since that contains the correct file context for /usr/libexec/pulpcore/gunicorn. However, it appears 1.1.2 didn't bump the versions in the .te files themselves so it'll likely fail to load. I pointed this out to @mikedep333 and he's working on an updated release.


ekohl:
This looks like an issue in Koji:
```
DEBUG util.py:598:  Getting requirements for pulpcore-selinux-1.1.3-1.el7.src
DEBUG util.py:598:   --> checkpolicy-2.5-8.el7.x86_64
DEBUG util.py:598:   --> selinux-policy-devel-3.13.1-266.el7.noarch
DEBUG util.py:598:  Error: Package: yum-3.4.3-167.el7.centos.noarch (build)
DEBUG util.py:598:             Requires: python-iniparse
```

mikedep333 (approved):

evgeni:
Yeah, Koji is irritated by the fact we have a python-iniparse SRPM which doesn't build a python-iniparse RPM :woman_shrugging: 

Needs untag, build, tag until I fix the source

evgeni:
[test rpm]

evgeni:
[test rpm]

evgeni:
Given python-pulpcore needs a bit of massage anyways when (re-generated) by pyp2rpm, I think it's fair to say we can also revert the wrappers change it will do.

ACK

evgeni (approved):

PR #43, opened 2020-10-09
Bump dynaconf Requires to skip RCs

evgeni:
pulpcore needs at least 3.0.0 final, otherwise distutils are unhappy


ekohl (approved):

ekohl:
I just realized we didn't use `Fixes #30963` to close https://projects.theforeman.org/issues/30963. Right now that's not set up, but the question is: should it be set up?

evgeni:
I don't really want to use redmine... ;)

ekohl:
That's fair, but then we should direct people to https://github.com/theforeman/pulpcore-packaging/issues if there is one.

PR #60, opened 2020-11-03
galaxy-ng 4.2.0 rc2

evgeni:


evgeni:
this needs pulp_ansible 0.5.0 which we don't have yet.

pcreech (approved):
This looks good

PR #62, opened 2020-11-05
fix license tags in rpm specfiles

evgeni:


packages/python-aiodns/python-aiodns.spec
```diff
@@ -3,10 +3,10 @@
 
 Name:           python-%{pypi_name}
 Version:        2.0.0
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Simple DNS resolver for asyncio
 
-License:        None
```
ekohl, line 9: https://github.com/saghul/aiodns/pull/84

packages/python-aiodns/python-aiodns.spec
```diff
@@ -3,10 +3,10 @@
...
+License:        MIT
 URL:            http://github.com/saghul/aiodns
```
ekohl, line 10: https://github.com/saghul/aiodns/pull/89

packages/python-backoff/python-backoff.spec
```diff
@@ -3,10 +3,10 @@
 
 Name:           python-%{pypi_name}
 Version:        1.10.0
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Function decoration for backoff and retry
 
-License:        None
```
ekohl, line 9: This one is weird because you'd think https://github.com/litl/backoff/blob/128bba5377568b8e2c0230973e887e0b4f5da800/pyproject.toml#L8 would be sufficient

packages/python-djangorestframework-queryfields/python-djangorestframework-queryfields.spec
```diff
@@ -3,10 +3,10 @@
 
 Name:           python-%{pypi_name}
 Version:        1.0.0
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Serialize a partial subset of fields in the API
 
-License:        None
```
ekohl, line 9: This is weird, it's in setup.py

packages/python-idna-ssl/python-idna-ssl.spec
```diff
@@ -3,10 +3,10 @@
 
 Name:           python-%{pypi_name}
 Version:        1.1.0
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Patch ssl.match_hostname for Unicode(idna) domains support
 
-License:        None
```
ekohl, line 9: Looks like this package is archived on Github so won't be fixed (nor will we see updates)

packages/python-jsonschema/python-jsonschema.spec
```diff
@@ -3,10 +3,10 @@
 
 Name:           python-%{pypi_name}
 Version:        3.2.0
-Release:        3%{?dist}
+Release:        4%{?dist}
 Summary:        An implementation of JSON Schema validation for Python
 
-License:        None
```
ekohl, line 9: It is in setup.cfg so that's odd.

packages/python-odfpy/python-odfpy.spec
```diff
@@ -3,10 +3,10 @@
 
 Name:           python-%{pypi_name}
 Version:        1.4.1
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Python API and tools to manipulate OpenDocument files
 
-License:        None
+License:        GPL-2.0-or-later or Apache-2.0
```
ekohl: I think the `-or-later` isn't really Fedora type. Shouldn't it be `GPL-2+`?

packages/python-pycares/python-pycares.spec
```diff
@@ -3,10 +3,10 @@
 
 Name:           python-%{pypi_name}
 Version:        3.1.1
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Python interface for c-ares
 
-License:        None
```
ekohl, line 9: https://github.com/saghul/pycares/pull/132

ekohl (commented):
For most I checked the upstream parts and submitted 2 PRs to fix them. I think only the Fedora code for GPL2+ is important here.

Spredzy (approved):
Thank you :+1:

packages/python-odfpy/python-odfpy.spec
evgeni: blah blah blah, can't we have a standard for writing license identifiers down? ;) (read: fixed)

ekohl: No, the world doesn't work that way. That would be too easy.

pcreech (approved):

ekohl (approved):

PR #69, opened 2020-11-27
reexec systemd after policy changes to make socket labeling work

evgeni:


packages/pulpcore-selinux/pulpcore-selinux.spec
```diff
@@ -79,6 +83,9 @@ fi
 
 
 %changelog
+* Fri Nov 27 2020 Evgeni Golov - 1.2.3-2
```
ekohl, line 86: Just 2 more till 1.2.3-4

ekohl (commented):
Would it be good to document that this is needed on systemd < 245?

evgeni:
[test rpm]

ekohl (approved):

PR #72, opened 2020-12-09
python-pulp-ansible 0.5.4 

evgeni:


evgeni:
includes #71, but I wanted individual PRs as I want individual builds for both releases

pcreech (approved):
ACK

PR #75, opened 2020-12-21
Correct deps

evgeni:


evgeni:
repoclosure complains about the pkg already in the repo, which this PR fixes.

ekohl (approved):
It is indeed only listed as extras.

PR #80, opened 2021-01-11
bump plugins

evgeni:


ekohl (commented):
Looks like CI doesn't understand epoch and is sad. I can understand that epochs make CI sad, but can we fix this?

evgeni:
> Looks like CI doesn't understand epoch and is sad. I can understand that epochs make CI sad, but can we fix this?

Huh? That was a legitimate mistake in the spec, fixed now.

pcreech (approved):
ACK

PR #87, opened 2021-01-25
Don't ship futurize and pasteurize in bin

evgeni:
We don't need the binaries, and they conflict with python2-future in
EPEL


evgeni:
@jlsherrill the cheapest I could come up with :)

jlsherrill (approved):
Works 4 me!

PR #95, opened 2021-02-22
galaxy-ng 4.2.2 and deps

evgeni:


packages/python-galaxy-importer/python-galaxy-importer.spec
```diff
@@ -33,9 +33,9 @@ Requires:       python%{python3_pkgversion}-ansible-builder < 1.0
 Requires:       python%{python3_pkgversion}-attrs < 21
 Requires:       python%{python3_pkgversion}-attrs >= 19.3.0
 Requires:       python%{python3_pkgversion}-bleach < 4
-Requires:       python%{python3_pkgversion}-bleach >= 3.1.3
-Requires:       python%{python3_pkgversion}-bleach-whitelist < 1
-Requires:       python%{python3_pkgversion}-bleach-whitelist >= 0.0.10
```
ekohl, line 38: Will you open a separate PR to remove the -whitelist package?

evgeni: We  keep the old plugins in the repo, and that still requires `-whitelist`, so no.

ekohl: Right, so that's something to clean up in 3.next. Isn't package renames "fun"?

ekohl (approved):

PR #102, opened 2021-03-03
Patches rq and pulpcore to heartbeat over 300s

bmbouter:
* pulpcore's worker timeout is adjusted to 300 seconds before it is
  considered missing
* rq is adjusted so that more heartbeats will occur during the worker
  timeout period

### DIFF
--- a/packages/python-pulpcore/0001-Adjusts-worker-timeout-to-300-seconds.patch
+++ b/packages/python-pulpcore/0001-Adjusts-worker-timeout-to-300-seconds.patch
@@ -1,0 +1,25 @@
+From 16d69707494bc9a0c70e49d9093c0b1b62deb58a Mon Sep 17 00:00:00 2001
+From: Brian Bouterse <bmbouter@redhat.com>
+Date: Wed, 3 Mar 2021 15:35:22 -0500
+Subject: [PATCH] Adjusts worker timeout to 300 seconds
+
+---
+ pulpcore/tasking/constants.py | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/pulpcore/tasking/constants.py b/pulpcore/tasking/constants.py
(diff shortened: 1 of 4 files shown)


dralley (approved):

evgeni:
ok to test

evgeni:
@bmbouter you need to bump the release in the header too ;)

otherwise this looks good

bmbouter:
@evgeni thanks! I updated the headers. </facepalm>

The pulpcore patch needs to be carried until `pulpcore>=3.11` at which point rpm based users can control this via a setting instead.

For the `rq` patch we're still waiting for the upstream to merge + release, so it's unclear how long that will be, but I'm hoping it'll be 1.7.1+  https://github.com/rq/rq/pull/1427

evgeni:
@bmbouter that means that pulpcore users *before* 3.11 will get different behaviour depending on whether they install from PyPi or RPM? That sounds odd… Wouldn't it be more correct to set the timeout to 300 in 3.9.2?

ACK @ rq -- do you want to hold the merge here until upstream comments on / accepts your PR, or are we confident enough to just ship it as is?

Packaging wise this is fine.

bmbouter:
@evgeni I see what you're saying. As a bit of background, the value isn't changing even in 3.9 or any other release actually, it's just configurable in 3.11. Since that configuration is a feature, we can't backport it to earlier version of Pulp anyway. Katello isn't in a position to upgrade to 3.11 so @jlsherrill requested the rpms receive this behavior by value adjustment. For PyPI users they can upgrade to 3.11 instead.

For RQ, I think it's ok to ship without waiting for the RQ merge. I asked this same question on IRC about waiting or not, and @jlsherrill confirmed also for us to go ahead. If for some reason it doesn't work out with upstream RQ we can adjust decide then, but I think it will.

evgeni (approved):

PR #127, opened 2021-03-31
fix ansible-lint requires

evgeni:


evgeni:
repoclosure is failing on the package *in* the repo, which this PR is fixing.

pcreech (approved):

PR #128, opened 2021-04-01
Bump python-galaxy-ng to 4.3.0a1

evgeni:


packages/python-galaxy-ng/python-galaxy-ng.spec
```diff
@@ -1,14 +1,16 @@
 # Created by pyp2rpm-3.3.3
 %global pypi_name galaxy-ng
 
+%global full_version %{version}a1
```
ekohl, line 4: Is it time for a `pre_release` macro as we have it in Foreman? I can imagine this would show up more often.

ekohl (approved):

packages/python-galaxy-ng/python-galaxy-ng.spec
evgeni: probably, yeah. I have a "do nightlies for some core packages", so I'd add that then.

ekohl: Fair enough

PR #139, opened 2021-04-14
galaxy-importer: Requires tar and manages /etc/galaxy-importer

Spredzy:


evgeni:
ok to test

Why should it manage a folder it ships no files in?

Spredzy:
The goal is to integrate as much of possible of this https://github.com/ansible/galaxy_collection/blob/devel/roles/post_install_config/tasks/galaxy_importer.yml in here.

> Why should it manage a folder it ships no files in?

So the folder is managed at the package level, and not by the orchestration tool. Orchestration tool will provide the file.
If this is really bad, we can provide a file, fully commented (but the header) with the various different option

evgeni:
I am not against the empty folder, just curious ;)

evgeni (approved):

PR #144, opened 2021-04-15
Update pulp_deb to 2.9.2 (pulpcore 3.7 repo)

quba42:
Needs to be coordinated with:
- https://github.com/theforeman/foreman-packaging/pull/6588
- https://github.com/Katello/katello/pull/9329
- The Katello 3.18.4 release


jlsherrill:
You're missing one step, you need to wget the tarball from pypi, git annex add it, and then git rm the old one.

quba42:
> You're missing one step, you need to wget the tarball from pypi, git annex add it, and then git rm the old one.

I am trying out the `obal` tooling right now. Looks like I am still missing some dependencies though.

quba42:
Manual `git annex` commands is easier after all than trying to get `obal` to work on a Debian host...

quba42:
Do not merge until I have figured out if this requires new API bindings on the Katello side!

quba42:
Updated the PR to update to 2.9.2 instead of 2.9.1. This is exactly the same bindings wise, but includes an important bugfix.

jturel:
[test rpm]

quba42:
Rebased to the tip of the `rpm/3.7` branch one more time.

evgeni (approved):

PR #151, opened 2021-04-29
add zchunk

evgeni:


evgeni:
fails on EL7, and I have no idea why (well, I maybe do, too old meson macros, or some shit)

evgeni:
seems we're fixing it in a different way.

PR #156, opened 2021-05-12
update requirements.txt to match reality

evgeni:


ekohl (approved):

ekohl:
I wonder if we could write a check that diffs the packages (spec files) and the version mentioned here so they must be in sync or CI fails.

PR #167, opened 2021-05-31
pulpcore 3.11.2

evgeni:


ekohl (approved):
I think I saw that the blocker regression in Django was fixed. Given that the dependency was updated, it looks ok :+1:

evgeni:
Correct, 3.11.2 has a fix to be compatible with django 2.2.23 :)

PR #169, opened 2021-06-01
Add postgresql-debversion-evr

m-bucher:
Adding the lightweight implementation of deb_version_compare function as PostgreSQL-extension.


m-bucher:
@evgeni Anything I can do about the build error?
```
error: line 5: Unknown tag: %scl_package postgresql-debversion-evr
```

evgeni:
my best guess is that the pulpcore build roots are not setup to use the postgresql SCL correctly.

which makes me wonder why we ship the package here, and not in katello packaging anyways.

@jlsherrill do you remember why we did it this way?

quba42:
> which makes me wonder why we ship the package here, and not in katello packaging anyways.

I think it is not just used for Katello, but also directly within `pulp_rpm` (the rpm extension that is), and we are planning to use this directly in `pulp_deb` as well, so `pulpcore-packaging` does make sense as a location.

jlsherrill:
i don't think the evr extension is used by pulp-rpm, we built it in the pulpcore tag simply because that was the easiest place to put it.  The existing katello/foreman tags use the tfm scl and iirc a tag can only build into a single scl.  

Most likely when we moved from foreman-packaging to pulpcore-packaging this scl configuration got lost, and i bet the evr extension hasn't been built since and will fail if we try to rebuild it.

So i see two paths forward (but there very well may be more):
1) create a new tag/build for these specific packages to build into the postgresql scl
2) use the postgresql scl for the existing tags/builds for all of el7 pulpcore the way we were doing before

and maybe the decision around python 3.7 on el7 will influence this decision

quba42:
> i don't think the evr extension is used by pulp-rpm

I had a look and the pulp_rpm extension has some version of the evr extension copied directly into a DB migration: https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/migrations/0013_RAW_rpm_evr_extension.py

So it is not using the evr extension as packaged in pulpcore-packaging but I am not sure this is the "nicest" way of doing things...

m-bucher:
@evgeni @jlsherrill any news here?
Should I rather add a PR in foreman-packaging?

packages/postgresql-debversion-evr/postgresql-debversion-evr.spec
```diff
@@ -0,0 +1,63 @@
...
+%build
+
+%install
+rm -rf $RPM_BUILD_ROOT
+%{?scl:scl enable %{scl} - <<EOF}
+%make_install
+%{?scl:EOF}
+
+%files
+%{_datadir}/pgsql/extension/debversion_evr--0.0.1.sql
```
ekohl:
```suggestion
%{_datadir}/pgsql/extension/debversion_evr--%{version}.sql
```

packages/postgresql-debversion-evr/postgresql-debversion-evr.spec
```diff
@@ -0,0 +1,63 @@
...
+%{?scl:Requires: %{?scl_prefix}runtime}
+
+%{?scl:BuildRequires: %{?scl_prefix}runtime}
+BuildRequires: %{?scl_prefix}postgresql-devel
+
+%if 0%{?rhel} > 7
+BuildRequires: postgresql-server-devel
+%endif
+
+ExclusiveArch: x86_64
```
ekohl, line 28: What makes this limited to this arch? Since it's pure SQL, isn't it noarch in the end?

packages/postgresql-debversion-evr/postgresql-debversion-evr.spec
```diff
@@ -0,0 +1,63 @@
+%if 0%{?rhel} <= 7
+%global scl rh-postgresql12
```
ekohl, line 2: Should we only do so if no SCL is already set? In some cases the buildroot should already set it, right?

packages/postgresql-debversion-evr/postgresql-debversion-evr.spec
m-bucher: Shamelessly copied form `postgresql-evr`'s SPEC :innocent: 
https://github.com/theforeman/pulpcore-packaging/blob/d081221df48e7c539d9ba9b4d742f237e40ae949/packages/postgresql-evr/postgresql-evr.spec#L28
Though I guess you are right about this.

packages/postgresql-debversion-evr/postgresql-debversion-evr.spec
ekohl: Setting the `scl` via a global doesn't really work as you can see in the build:
```
error: line 5: Unknown tag: %scl_package postgresql-debversion-evr
```

You already need the SCL tools in the buildroot. There's a good chance Koji wasn't set up for that.

jlsherrill:
@m-bucher and i discussed this a bit and recently there was some discussion about moving these postgres extensions into the katello code base as part of a db migration.  I opened this pr to help facilitate that:  https://github.com/theforeman/foreman/pull/8660/files

If that goes well, that effort could replace this and an rpm wouldn't need to be built.

m-bucher:
Closing this in favour of https://github.com/Katello/katello/pull/9507

PR #178, opened 2021-06-17
place pulpcore-worker in libexec

evgeni:


packages/python-pulpcore/python-pulpcore.spec
```diff
@@ -166,17 +166,21 @@ do
   install -D -m 755 ${wrapper} %{buildroot}%{_libexecdir}/%{pypi_name}/${wrapper}
 done
 
+mv %{buildroot}%{_bindir}/pulpcore-worker %{buildroot}%{_libexecdir}/%{pypi_name}/
```
ekohl: Should we use `install -D -m 755` to ensure permissions?

evgeni:
needs to wait for https://pulp.plan.io/issues/8906 to be decided

packages/python-pulpcore/python-pulpcore.spec
evgeni: it is in bindir, so I would expect it to already have correct perms.

evgeni: I went with wrapper for consistency

ekohl (approved):

PR #184, opened 2021-07-01
galaxy-ng 4.2.4

evgeni:


packages/python-galaxy-ng/python-galaxy-ng.spec
```diff
@@ -41,6 +41,10 @@ Requires:       python%{python3_pkgversion}-setuptools
 # Remove bundled egg-info
 rm -rf %{pypi_name}.egg-info
 
+# drop the ansible requirement, we don't ship ansible for Python3
```
ekohl, line 44: Is there an upstream issue for this?

evgeni: no

pcreech (approved):
ACK from me.  Not sure if comment by @ekohl should be blocking

evgeni:
> ACK from me. Not sure if comment by @ekohl should be blocking

I think we cleared that on IRC ;)

PR #203, opened 2021-07-29
Update pulp-ansible to 0.9.0

Odilhao:


jlsherrill:
i'd adjust the pulpcore requires to match here: https://github.com/pulp/pulp_ansible/blob/0.9/requirements.txt#L5

also i think we like to keep https://github.com/theforeman/pulpcore-packaging/blob/rpm/3.14/automation/requirements.txt up to date

Odilhao:
> i'd adjust the pulpcore requires to match here: https://github.com/pulp/pulp_ansible/blob/0.9/requirements.txt#L5
> 
> also i think we like to keep https://github.com/theforeman/pulpcore-packaging/blob/rpm/3.14/automation/requirements.txt up to date


Done, I forgot that we have pulp-ansible inside the requirements.txt

---

Also changed from `pulpcore >= 3.12.1` to `pulpcore >= 3.14`

jlsherrill (approved):

PR #207, opened 2021-08-10
Add mock configs

ehelms:
@evgeni @Odilhao This adds mock configs for EL7 nonscl, SCL and EL8. This should provide a starting place to be able to update python specs to have scl_prefix options in them and then test builds. The syntax using obal is:

```
obal mock <package> --config mock/el7-scl.cfg
```

### DIFF
--- a/mock/el7-nonscl.cfg
+++ b/mock/el7-nonscl.cfg
@@ -1,0 +1,41 @@
+config_opts['root'] = 'pulpcore-el7-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
+config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils findutils gawk gcc gcc-c++ grep gzip info make patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
+config_opts['chroot_additional_packages'] = 'yum vim'
+config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
+# beware RHEL use 7Server or 7Client
+config_opts['releasever'] = '7'
+config_opts['package_manager'] = 'yum'
+
+config_opts['yum.conf'] = """
+[main]
+cachedir=/var/cache/yum
+debuglevel=1
+reposdir=/dev/null
+logfile=/var/log/yum.log
+retries=20
+obsoletes=1
+gpgcheck=0
+assumeyes=1
+syslog_ident=mock
+syslog_device=
+
+# repos
+[base]
+name=BaseOS
+enabled=1
+mirrorlist=http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os
+failovermethod=priority
+
+[updates]
+name=updates
+enabled=1
+mirrorlist=http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=updates
+failovermethod=priority
+
+[pulpcore]
+name=pulpcore
+enabled=1
+baseurl=https://yum.theforeman.org/pulpcore/3.14/el7/x86_64/
+"""
--- a/mock/el7-nonscl.cfg
+++ b/mock/el7-nonscl.cfg
@@ -37,5 +37,5 @@ failovermethod=priority
 [pulpcore]
 name=pulpcore
 enabled=1
-baseurl=https://yum.theforeman.org/pulpcore/3.14/el7/x86_64/
+baseurl=https://yum.theforeman.org/pulpcore/3.15/el7/x86_64/
 """
--- a/mock/el7-scl.cfg
+++ b/mock/el7-scl.cfg
@@ -40,5 +40,5 @@ baseurl=http://mirror.centos.org/centos/7/sclo/x86_64/rh/
 
 [pulpcore]
 name=pulpcore
-baseurl=https://yum.theforeman.org/pulpcore/3.14/el7/x86_64/
+baseurl=https://yum.theforeman.org/pulpcore/3.15/el7/x86_64/
 """
--- a/mock/el7-scl.cfg
+++ b/mock/el7-scl.cfg
@@ -1,0 +1,44 @@
+config_opts['root'] = 'pulpcore-scl-el7-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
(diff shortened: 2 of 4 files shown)


evgeni:
One thing that I forgot when we talked earlier: for EL8 we will need to restructure our flattened modules, so that we can pick py38 there too (right now it's impossible).

@pcreech has a todo item for this somewhere, but wanted to raise here to list all known blockers.

ehelms:
Looks like it already is? http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001-mod/python38:3.8/

evgeni:
Mumbling to myself:
* ~~To build against Python 3.8 on EL8, we need to install `python38-rpm-macros` which will take care to update `python3_pkgversion` macro and then everything "just" works (verified in a podman container), but I have no idea where to add that to `el8.cfg`?! Adding that to `config_opts['chroot_setup_cmd']` or `config_opts['chroot_additional_packages']` doesn't seem to change anything?~~ Latest changes from @ehelms take care of that.
* On EL7 nothing builds, as the SCL packages are named `rh-python38-python-devel` and not `rh-python38-python3-devel`, thus breaking our `BuildRequires`…
* I can get it to build on EL7 by dropping all use of `python3_pkgversion`, but this will break EL8…

evgeni:
With EL8 sorted with the latest changes, let's look at EL7:
One can convert an RPM spec with `spec2scl` easily, and it looks sane afterwards. But…

As there is nothing in the SCL defining `python3_pkgversion`, the non-SCL value (`3`) gets picked up, which translates things like `%{?scl_prefix}python%{python3_pkgversion}-devel` to `rh-python38-python3-devel`. However, the SCL decided not to actually use that name and the real package is called `rh-python38-python-devel`.
The wise packager says "cool, let's empty that macro" -- `%{?scl:%global python3_pkgversion %{nil}}` and off we go.
But not too far, as this now means that both the source and the binary package is called `rh-python38-python-foo` which blows up our SPEC: `error: line 26: %package -n     python-aiohttp: package python-aiohttp already exists` as we have explicit package sections for the binary packages, as we default to having `python-foo.srpm` and `python3-foo.rpm`.

evgeni:
I think the cleanest way would be:
* Define an own SCL `tfm-py38` (or whatever name)
* In that SCL, packages would remain `%{?scl_prefix}python%{python3_pkgname}-%{pypi_name}`, so `tfm-py38-python3-aiohttp`
* Dependencies outside the SCL would need to not use `python3_pkgname`, so our specs will need something like:
  ```
  %{?scl:%global scl_python3_pkgname %{nil}}
  %{!?scl:%global scl_python3_pkgname %{python3_pkgname}}
  ```
  And then using `%{scl_python3_pkgname}` on those deps…

ehelms:
> I think the cleanest way would be:
> 
> * Define an own SCL `tfm-py38` (or whatever name)
> * In that SCL, packages would remain `%{?scl_prefix}python%{python3_pkgname}-%{pypi_name}`, so `tfm-py38-python3-aiohttp`
> * Dependencies outside the SCL would need to not use `python3_pkgname`, so our specs will need something like:
>   ```
>   %{?scl:%global scl_python3_pkgname %{nil}}
>   %{!?scl:%global scl_python3_pkgname %{python3_pkgname}}
>   ```
>     
>   And then using `%{scl_python3_pkgname}` on those deps…

Based on discussions, I believe we decided this was the cleanest way given the constraints.

evgeni:
Call me crazy, but what if, and hear me out: we proivide an own, properly named `tfm-python38-python3-(devel|setuptools)` that just depends on the "real" thing from `rh-python38`?

NO MORE WEIRD SPECIALCASES

ehelms:
> Call me crazy, but what if, and hear me out: we proivide an own, properly named `tfm-python38-python3-(devel|setuptools)` that just depends on the "real" thing from `rh-python38`?

`-devel` comes from the Python package, so you'd basically be building python and that feels ugly. 

Since we can control the spec template, another option, which I guess is a bit repetitive is to define two sets of build and requires? e.g.

```
if %{?scl}
BuildRequires:  %{?scl_prefix_python}python-devel
BuildRequires:  %{?scl_prefix_python}python-setuptools
%else
BuildRequires:  %{?scl_prefix_python}python%{?python3_pkgversion}-devel
BuildRequires:  %{?scl_prefix_python}python%{?python3_pkgversion}-setuptools
%endif
```

We would only have to do this for packages coming from the `rh-python38` SCL that we require.

evgeni:
Oh no, I wouldn't be *building* Python. Just providing an empty "meta" package pulling in the "real" rh-python38-python-devel.

evgeni:
see #212 for an actual example :)

mock/el7-nonscl.cfg
```diff
@@ -0,0 +1,41 @@
...
+[updates]
+name=updates
+enabled=1
+mirrorlist=http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=updates
+failovermethod=priority
+
+[pulpcore]
+name=pulpcore
+enabled=1
+baseurl=https://yum.theforeman.org/pulpcore/3.14/el7/x86_64/
```
evgeni:
```suggestion
baseurl=https://yum.theforeman.org/pulpcore/3.15/el7/x86_64/
```

mock/el7-scl.cfg
```diff
@@ -0,0 +1,44 @@
...
+mirrorlist=http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=updates
+failovermethod=priority
+
+[sclo-rh]
+name=sclo-rh
+baseurl=http://mirror.centos.org/centos/7/sclo/x86_64/rh/
+
+[pulpcore]
+name=pulpcore
+baseurl=https://yum.theforeman.org/pulpcore/3.14/el7/x86_64/
```
evgeni:
```suggestion
baseurl=https://yum.theforeman.org/pulpcore/3.15/el7/x86_64/
```

mock/el8.cfg
```diff
@@ -0,0 +1,67 @@
...
+module_hotfixes=1
+
+[koji-python-38]
+name=koji-python-38
+baseurl=http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001-mod/python38:3.8/
+module_hotfixes=1
+
+[pulpcore]
+name=pulpcore
+baseurl=https://yum.theforeman.org/pulpcore/3.14/el8/x86_64/
```
evgeni:
```suggestion
baseurl=https://yum.theforeman.org/pulpcore/3.15/el8/x86_64/
```

evgeni (approved):

PR #212, opened 2021-08-18
add py38 scl 

evgeni:


evgeni:
the SCL package in this PR, allowed me, after installing the following "deps" (and their deps, obviously) in my buildroot, to produce a working scl'ed package:
```
rh-python38-python-rpm-macros-3.8.6-1.el7.noarch
rh-python38-python-srpm-macros-3.8.6-1.el7.noarch
tfm-python38-build-1.0-1.el7.x86_64
```

packages/tfm-python38/tfm-python38.spec
```diff
@@ -0,0 +1,144 @@
...
+
+cat >> %{buildroot}%{_scl_scripts}/enable << EOF
+. scl_source enable %{scl_python}
+EOF
+
+# Create the scldevel subpackage macros
+
+cat >> %{buildroot}%{_root_sysconfdir}/rpm/macros.%{scl_name_base}-scldevel << EOF
+%%scl_%{scl_name_base} %{scl}
+%%scl_prefix_%{scl_name_base} %{scl_prefix}
```
evgeni: this produces warnings:
```
warning: Macro %scl_tfm needs whitespace before body
warning: Macro %scl_prefix_tfm needs whitespace before body
```

seems it doesn't like the `-` in the `scl_name_base`

comps/comps-pulpcore-el7.xml
```diff
@@ -163,6 +163,12 @@
       <packagereq type="default">rh-postgresql12-postgresql-debversion</packagereq>
       <packagereq type="default">rh-postgresql12-postgresql-evr</packagereq>
       <packagereq type="default">ruby-solv</packagereq>
+      <packagereq type="default">tfm-python38</packagereq>
```
ehelms: I was curious why name it a python SCL vs. a Pulp SCL?

packages/tfm-python38/README
```diff
@@ -0,0 +1,29 @@
+Package %{scl_name} provides dependencies for the Foreman application,
+shipped as a SoftwareCollection. For more information about Software
+Collections, see the scl(1) man page. By installing the %{scl_name}
+collection, you will get the minimum working set of packages to install
+Foreman.  This collection is not supported for other purposes.
+
+Usage: scl enable %{scl} 'ruby'
```
ehelms: `python`

packages/tfm-python38/tfm-python38.spec
```diff
@@ -0,0 +1,144 @@
...
+
+%package scldevel
+Summary: Package shipping development files for %scl
+
+%description scldevel
+Package shipping development files, especially useful for development of
+packages depending on %scl Software Collection.
+
+%package python3-devel
+Summary: Package pulling in %{scl_prefix_python}python-devel
```
ehelms: Maybe something like `empty package to pull in....` or include somewhere in the description a note about *why* we have to do this? Thinking of things to help us remember or inform any othres that come looking at this code.

packages/tfm-python38/tfm-python38.spec
```diff
@@ -0,0 +1,144 @@
...
+%%scl_prefix_%{scl_name_base} %{scl_prefix}
+EOF
+
+
+# install generated man page
+mkdir -p %{buildroot}%{_mandir}/man7/
+install -m 644 %{scl_name}.7 %{buildroot}%{_mandir}/man7/%{scl_name}.7
+
+%files
+%if 0%{?rhel} <= 6
```
ehelms: You could probably drop this :)

comps/comps-pulpcore-el7.xml
evgeni: I don't mind either way.
My trail of thought was: we're providing Python to run Pulp in, so Python.

packages/tfm-python38/tfm-python38.spec
evgeni: Probably. Something something templates ;)

comps/comps-pulpcore-el7.xml
ehelms: Will the Pulp packages also be in this SCL?

packages/tfm-python38/tfm-python38.spec
```diff
@@ -0,0 +1,143 @@
+%global __python /usr/bin/python3
+%global scl_name_base tfm-python
+%global scl_name_version 38
```
ehelms: I wonder, given the warning below, if this should be broken out by:

```
%global scl_name_prefix tfm-
%global scl_name_base python
%global scl_name_version 38
```

comps/comps-pulpcore-el7.xml
evgeni: Yeah

ehelms: So then I think I owuld lean towards either:

```
tfm-pulpcore
pulpcore-python38
```

I think without the pulp name in there, folks looking at the packages and debugging via `scl enable` will get confused and we'll field more questions than we want.

evgeni: I kinda dislike the `pulpcore` prefix, as this (for me) implies that the SCL is somehow related to the Pulp *Project*, which it isn't (and shouldn't be).

Onother thing that came to my mind when you asked about the prefix yesterday: what will/should happen if we move to Python 3.10 (or whatever)? For Foreman, we've been keeping the `tfm` SCL while switching Rubies and Nodes underneath. And while this is nice for the "you can just install `tfm-<whatever>` without thinking which Ruby version it uses", it always struck me a bit odd as verifying a change in the underlying stack always required looking at the actual dependencies of the built RPMs. But then again it's really nice for the installer and friends…

Voting `tfm-pulpcore` then.

What about you?

evgeni:
Meh, this somehow still installs the actual files to `/opt/rh/rh-python38/` :/

Digging more.

comps/comps-pulpcore-el7.xml
ekohl: It also means you can `yum update` and upgrade. If we'd introduce a version in the SCL name then that becomes much more complicated. I'm not sure the packaging templates have `Obsoletes` and obsoleting something from another SCL feels weird.

:+1: for `tfm-pulpcore`.

packages/tfm-pulpcore/tfm-pulpcore.spec
```diff
@@ -0,0 +1,184 @@
...
+%%global py3_build_egg %%python38py3_build_egg
+%%global py3_build_wheel %%python38py3_build_wheel
+%%global py3_install_egg %%python38py3_install_egg
+%%global py3_install_wheel() %%%%{expand:%%{python38py3_install_wheel_start}%%%%{1}%%{python38py3_install_wheel_end}}
+%%global python3_sitelib %pulpcore_sitelib
+%%global python3_sitearch %pulpcore_sitearch
+%%global pulpcorepy3_install() %%{expand:\\\\\\
+  %%{?scl:scl enable rh-python38 '}\\\\\\
+  CFLAGS="\${CFLAGS:-\${RPM_OPT_FLAGS}}" LDFLAGS="\${LDFLAGS:-\${RPM_LD_FLAGS}}"\\\\\\
+  %%{python38__python3} %%{python38py_setup} %%{?py_setup_args} install -O1 --skip-build --root %%{buildroot} --install-purelib %%{python3_sitelib} --install-platlib %%{python3_sitearch} --install-scripts %%{_bindir} %%{?*}\\\\\\
```
evgeni, line 145: I had to add `--install-purelib %%{python3_sitelib} --install-platlib %%{python3_sitearch} --install-scripts %%{_bindir}` here, as otherwise stuff would land in the respective `rh-python38` folders.

For the curious reader: I *wanted* to just add that to the `%python38py3_install` macro, but failed as the following didn't work:
```
 %%global py3_install %%python38py3_install --install-purelib …
```
It complained that `-` is not a valid param for the macro, which is true, I wanted to add it to the `%{?*}` that the original macro has… But how?

evgeni:
Note to self, the libexec wrappers in `python-pulpcore` will require some trivial adaption (untested):

https://github.com/theforeman/pulpcore-packaging/blob/2b69c4af86d45833d3530ba9e78db301fc25abe5/packages/python-pulpcore/python-pulpcore.spec#L158-L161

```
for wrapper in %{wrappers}
do
  printf '#!/bin/bash\nexec %{?scl:scl enable tfm-pulpcore --} %s "$@"\n' ${wrapper} > ${wrapper}
done
```

evgeni:
This could use a final review :)

pcreech (approved):
ACK

PR #213, opened 2021-08-18
Update python-pulpcore to 3.14.4

Odilhao:


evgeni:
you'll have to update `aioredis` package, as we currently package 2.0.0b1, but this now requires 2.0.0:
https://github.com/pulp/pulpcore/compare/3.14.3...3.14.4#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552

evgeni (approved):

PR #215, opened 2021-08-24
Depend on mailcap for mime.types

ekohl:
aiohttp imports stdlib's mimetypes which attempts to read various locations where mime.types can be. In Python 3.6 this is:

```python
knownfiles = [
    "/etc/mime.types",
    "/etc/httpd/mime.types",                    # Mac OS X
    "/etc/httpd/conf/mime.types",               # Apache
    "/etc/apache/mime.types",                   # Apache 1
    "/etc/apache2/mime.types",                  # Apache 2
    "/usr/local/etc/httpd/conf/mime.types",
    "/usr/local/lib/netscape/mime.types",
    "/usr/local/etc/httpd/conf/mime.types",     # Apache 1.2
    "/usr/local/etc/mime.types",                # Apache 1.3
    ]
```

Until a file is found, the next location is tried. On EL7 and EL8 the mailcap package provides `/etc/mime.types`.

This also solves a SELinux denial where `pulpcore_server_t` tries to read files from `/etc/httpd`. Since it's lower on the list than `/etc/mime.types`, this should be resolved.

This was found in https://github.com/pulp/pulpcore-selinux/pull/37 and has more information.

### DIFF
--- a/packages/python-aiohttp/python-aiohttp.spec
+++ b/packages/python-aiohttp/python-aiohttp.spec
@@ -3,7 +3,7 @@
 
 Name:           python-%{pypi_name}
 Version:        3.7.4
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Async http client/server framework (asyncio)
 
 License:        Apache 2
@@ -32,6 +32,9 @@ Requires:       python%{python3_pkgversion}-typing-extensions >= 3.6.5
 Requires:       python%{python3_pkgversion}-yarl < 2.0
 Requires:       python%{python3_pkgversion}-yarl >= 1.0
 
+# aiohttp depends on stdlib's mimetypes which reads /etc/mime.types
+Requires:       /etc/mime.types
+
 %description -n python%{python3_pkgversion}-%{pypi_name}
 %{summary}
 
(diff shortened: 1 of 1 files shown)


ekohl:
I will admit this is purely theoretical. I haven't actually tried to see if this matters. It also looks like `httpd` itself already depends on `/etc/mime.types`.

```console
# rpm -q --requires httpd      
/etc/mime.types
...
```

Should we do the same here?

evgeni (commented):
I'm happy with either package name or filename.

Just had to laugh about `/usr/local/lib/netscape/mime.types`, Netscape is a name I've not seen in a while.

ekohl:
Rebased to depend on the file directly, just like httpd does. I've also opened https://github.com/theforeman/pulpcore-packaging/pull/280 for 3.15.

packages/python-aiohttp/python-aiohttp.spec
```diff
@@ -3,7 +3,7 @@
 
 Name:           python-%{pypi_name}
 Version:        3.7.4
-Release:        1%{?dist}
+Release:        2%{?dist}
```
evgeni, line 6: This won't work, as 3.7.4-2 already exists (and is the py38 rebuild)

We could version this 3.7.4-1.1?

ekohl: The alternative is to leave this fix out of 3.14. In Foreman it doesn't really matter since Apache depends on the same file and we don't support running without Apache.

evgeni: works for me!

PR #218, opened 2021-08-25
2to3 migration 0.11.3

evgeni:


ekohl (approved):

quba42 (approved):
Apparently this was first bumped to 0.11.1 by different people for different branches.
Is there any policy on how to handle this sort of thing? I am guessing the preferred way is to bump on the newest available branch first, and then backport/cherry-pick from there?

evgeni:
> Is there any policy on how to handle this sort of thing? I am guessing the preferred way is to bump on the newest available branch first, and then backport/cherry-pick from there?

Correct. But given the diff is only the changelog, we don't care too much ;)

PR #225, opened 2021-09-01
prepare pulpcore 3.15 config

evgeni:


Odilhao:
Some files here have the same content as the on #212 , should we not consider the draft?

evgeni:
I'll rebase #212 once this is merged.

pcreech:
ACK

PR #228, opened 2021-09-03
SCL-ize gnupg

evgeni:


evgeni:
besides the obviously failing repoclosure, this seems to work as it should. :metal:

pcreech (approved):
ACK

PR #231, opened 2021-09-06
add azure dependencies for django-storages

evgeni:


zjhuntin:
Blergh meant to approve before merge, but ACK

PR #235, opened 2021-09-06
Python 3.8 rebuild: A-C

evgeni:


packages/python-aiodns/python-aiodns.spec
```diff
@@ -1,56 +1,79 @@
...
+
 * Fri Jun 11 2021 Evgeni Golov 3.0.0-1
 - Update to 3.0.0
 
 * Thu Nov 05 2020 Evgeni Golov - 2.0.0-3
 - Fix License tag in spec file
 
 * Wed Apr 01 2020 Evgeni Golov - 2.0.0-2
-- Add python3-typing to Requires
+- Add python%{python3_pkgversion}-typing to Requires
```
evgeni, line 76: Well, okay, *maybe* the `sed` was a bit too broad, but hey!

Odilhao (approved):

PR #241, opened 2021-09-08
Python 3.8 rebuild: stage two, part two

evgeni:


evgeni:
You guessed it, repoclosure is still totally irrelevant :woman_shrugging:

pcreech (approved):

PR #243, opened 2021-09-08
Python 3.8 rebuild: stage three

evgeni:
needs #240 and #241 ;)


evgeni:
[test rpm]

pcreech (approved):
ACK pending pipe (repoclosure fail expected)

evgeni:
I dropped yarl from this PR, as it needs typing-extensions, which is in this PR…

evgeni:
[test rpm]

evgeni:
[test rpm]

evgeni:
same with aioredis, because I can't grep

PR #244, opened 2021-09-08
Python 3.8 rebuild: stage four

evgeni:
Needs #243


pcreech (approved):
ACK pending pipe (repoclosure fail expected)

evgeni:
[test rpm]

PR #249, opened 2021-09-09
correct drf provides to mention py 3.8

evgeni:


evgeni:
trivial fix is trivial

pcreech (commented):
retro ACK

PR #251, opened 2021-09-09
pulpcore 3.15

evgeni:


evgeni:
[test rpm]

evgeni:
i've split out the plugins to #254 as they are steight forward, while this is not

packages/python-pulpcore/python-pulpcore.spec
```diff
@@ -147,38 +103,58 @@ Using Pulp you can:
...
+  install -D -m 755 ${wrapper} %{buildroot}%{_root_libexecdir}/%{pypi_name}/${wrapper}
 done
 
-%files -n python%{python3_pkgversion}-%{pypi_name}
+%files -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 %license LICENSE
 %doc README.md
 %{_bindir}/pulp-content
 %{_bindir}/pulpcore-manager
 %{_bindir}/pulpcore-worker
```
evgeni, line 168: This puts these into the SCL'ed bindir.

For the worker, we use the libexec wrapper anyways, so it shouldn't hurt.
`pulpcore-manager` is called by `puppet-pulpcore` -- should we place a wrapper in `/usr/bin` that does the SCL dance for us?
And no idea where/how `pulp-content` is used.

CC @ekohl

evgeni: Pulp's Ansible-based installer also looks for a `/usr/bin/pulpcore-manager`, so I guess a wrapper is in order.

ekohl: :+1: for a wrapper that activates the SCL if needed. That makes it easy for the installer(s) and users.

I'd also love it if `pulpcore-manager` on packaging actually set a default for `PULP_SETTINGS` but that's probably out of scope.

evgeni: added.

pcreech (approved):
ACK

PR #254, opened 2021-09-10
plugins for 3.15

evgeni:
this is missing certguard and rpm, which will be in an own PR


evgeni:
it builds, ship it :ship:

pcreech (commented):
Retro ACK

PR #256, opened 2021-09-10
Update typing

evgeni:


evgeni:
[test rpm]

evgeni:
nope, doesn't work.

PR #258, opened 2021-09-10
Add python-parsley package

evgeni:


Odilhao (approved):

Odilhao:
I think that repoclosure failing is expected for now

PR #262, opened 2021-09-13
Rebuild ansible-lint against Python 3.8

evgeni:


evgeni:
closure :see_no_evil:

PR #269, opened 2021-09-13
build libcomps for python 3.8

evgeni:


evgeni:
@pcreech @Odilhao mind having a look?

doesn't build yet, but my idea is to build most of the package outside the SCL, and only the Python3 bits into the SCL, as I don't want to SCL'ize the shared library part, as it seems crazy?

pcreech:
Agreed.  Should only do the python package part in scl.  Fun times ahead....

evgeni:
```
Building for python3
-- Found PythonInterp: /opt/rh/rh-python38/root/usr/bin/python3 (found suitable version "3.8.11", minimum required is "3") 
-- Found PythonLibs: /usr/lib64/libpython2.7.so (found version "2.7.5") 
```

Yeah, sure :disappointed:

evgeni:
So I must admit I am stuck.

I can get it to build, but then either *everything* is inside the SCL (including `libcomps.so`, when using the `scl_pacakge` macro), or only the Python 3 bits are inside, but then the Requirements of the `tfm-pulpcore-python3-libcomps` pacakge are wrong (they ask for `python-abi(3.8)`, not `rh-python38-python(abi) = 3.8`)

Help?

evgeni:
Okay, I think this is working now…

pcreech (approved):
Everything looks good, from what I can tell.  ACK!

PR #270, opened 2021-09-15
Changes to make libcomps land in /usr/lib

pcreech:


pcreech:
@evgeni Here's as far as I've gotten.  The only thing is libcomps requires tfm-pulpcore-runtime, i'm not sure if we're going to get around that one.

For the life of me, i couldn't get the includes to land in /usr/include.  that's the DINCLUDE_INSTALL_DIR mess in the arguments.

It fails to build on EL8, we'll need to have a strategy for that, as _root{blah} macros don't exist outside of the scl.  perhaps exporting a common variable we set conditionally.

But basically, the strategy is to provide all the non-scl variable values (_root{blah}) as arguments to the cmake command, except for python scl related values.

Feel free to steal however you'd like, to pick up the torch.

pcreech:
actually, the include files are just header files anyways............ 

So we can probably just move them. (sfter thought, not enough time to implement at the moment)

PR #272, opened 2021-09-15
Add fips compliency patch to django

mdellweg:
The patch is taken from:

https://github.com/django/django/pull/14763


evgeni:
I think you need to add -p1 to the autosetup call, as it defaults to -p0 (which is silly)

mdellweg:
It is green now.
@evgeni I was told, packaging needs this ASAP. Do you need any sign off from the pulp team about this patch?

https://pulp.plan.io/issues/9355

evgeni:
Thanks. You opening the PR is good enough for me ;)

PR #286, opened 2021-09-29
obsolete python 3.6 packages that clash on binary names

evgeni:


evgeni:
I am kinda sure this is not all that we need, but it's a start.

pcreech (approved):

packages/python-pulp-cli/python-pulp-cli.spec
```diff
@@ -39,6 +39,9 @@ Requires:       %{?scl_prefix}python%{python3_pkgversion}-requests >= 2.24
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-schema = 0.7.4
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-setuptools
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-toml = 0.10.2
+%if 0%{?!scl:1}
+Obsoletes:      python3-%{pypi_name} < %{version}-%{release}
+%endif
```
m-bucher, line 44: I just hit that on EL7:
```
Transaction check error:
  file /usr/bin/pulp from install of tfm-pulpcore-python3-pulp-cli-0.14.0-1.el7.noarch conflicts with file from package python3-pulp-cli-0.11.0-1.el7.noarch
```
Should this `Obsoletes` not always be added?

Odilhao: We probably need to remove the obsolete block and add a exclude one for `/usr/bin/pulp` 

Like here https://github.com/theforeman/pulpcore-packaging/commit/980bd719d13481f4999847ecaefdb5eb5dd34779

Are you upgrading from Pulp 3.14 to 3.15 @m-bucher ?

m-bucher: @Odilhao 3.14 to 3.16

My idea would have been to remove the `%if` makro around the `Obsoletes` so `tfm-pulpcore-python3-pulp-cli-x.y.z` will also obsolete `python3-pulp-cli`-packages with version smaller than x.y.z

Odilhao: Just created #542 let's work there to solve the issue

PR #288, opened 2021-09-29
Adds aiohttp secure proxy support

bmbouter:


evgeni:
ok to test

The version/release will clash with the one in 3.15, we'd need to set release to 1.1 or so, instead of 2

bmbouter:
@evgeni like this?

evgeni:
Yeah, looks good. Will have a better look tomorrow, but on first glance this looks exactly as it should

evgeni (approved):

PR #296, opened 2021-10-13
build libcomps against python 3.6 AND 3.8 on EL8

evgeni:


packages/libcomps/libcomps.spec
```diff
@@ -102,6 +110,18 @@ Obsoletes:      %{?scl_prefix}platform-python-%{name} < %{version}-%{release}
 Python3 bindings for libcomps library.
 %endif
 
+%if %{with python36}
+%package -n python3-%{name}
```
evgeni, line 114: I named the package `python3-libcomps`, and not `python36-libcomps`, to avoid a horrible `Obsoletes` dance with the original `python3-libcomps` in EL8

evgeni:
requires https://github.com/theforeman/tool_belt/pull/406 to build. but  took the liberty to execute that already

jlsherrill (approved):

PR #298, opened 2021-10-18
add provides for pulpcore-plugin()

evgeni:
and obsolete old (python 3.6) names

### DIFF
--- a/packages/python-pulp-certguard/python-pulp-certguard.spec
+++ b/packages/python-pulp-certguard/python-pulp-certguard.spec
@@ -6,7 +6,7 @@
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        1.5.0
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Certguard plugin for the Pulp Project
 
 License:        GPLv2+
@@ -31,6 +31,8 @@ Requires:       %{?scl_prefix}python%{python3_pkgversion}-pyOpenSSL
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-pulpcore >= 3.10
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-setuptools
 
+Provides:       pulpcore-plugin(certguard) = %{version}
+Obsoletes:      python3-%{pypi_name} < %{version}-%{release}
 
 %description -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 A Pulp plugin that provides an X.509 capable ContentGuard for pulpcore.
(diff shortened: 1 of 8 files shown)


packages/python-pulp-certguard/python-pulp-certguard.spec
```diff
@@ -31,6 +31,8 @@ Requires:       %{?scl_prefix}python%{python3_pkgversion}-pyOpenSSL
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-pulpcore >= 3.10
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-setuptools
 
+Provides:       pulpcore-plugin(certguard) = %{veru}
```
ekohl:
```suggestion
Provides:       pulpcore-plugin(certguard) = %{version}
```

ekohl (commented):
:+1: for the general direction. Looks like there's one typo though.

packages/python-pulp-certguard/python-pulp-certguard.spec
evgeni: lol, vim, what did you do?!

ekohl (commented):
Can we have this in 3.14 too so we can start to change the installer?

Thinking more about this: I wonder if the installer can ensure `pulpcore-plugin(galaxy-ng)` to be present or if Puppet can't do that. That'd be a good thing to verify.

evgeni:
Yes, I intend to backport that to 3.14 once everyone agreed on the color of the bikeshed.

I think modern Puppet understands provides.

ekohl:
I tried:
```puppet
package { 'tfm-rubygem(foreman_puppet)':
  ensure => installed,
}
```

Then this results in:
```console
# puppet --version
7.11.0
# puppet apply --noop --debug test.pp 
....
Info: Applying configuration version '1634568878'
Debug: Prefetching yum resources for package
Debug: Executing: '/bin/rpm --version'
Debug: Executing '/bin/rpm -qa --nosignature --nodigest --qf '%{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n' | sort'
Debug: Executing: '/bin/rpm -q tfm-rubygem(foreman_puppet) --nosignature --nodigest --qf %{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n'
Debug: Finishing transaction 12300
Debug: Storing state
Debug: Pruned old state cache entries in 0.00 seconds
Debug: Stored state in 0.21 seconds
Notice: Applied catalog in 0.92 seconds
```

Looks like it uses `rpm -q $PACKAGE` so I think that confirms it should work.

ekohl (approved):

PR #300, opened 2021-10-19
add provides for pulpcore-plugin()

evgeni:
(cherry picked from commit 23cdd457c35473efb2d93d7684f6b5ec422e7225)


ekohl (approved):

evgeni:
Right, galaxy-ng is not compatible with 3.14, and given I don't intend to backport that any further, I'll just drop that change here.

ekohl (approved):

PR #304, opened 2021-10-20
Add provides for "pulpcore"

evgeni:


ekohl (approved):
I still wonder if at some point we should introduce a real `pulpcore` package that depends on the python3-pulpcore package and contains the systemd services + user creation. Would this provides make that transition harder in the future? I don't think so, but I want to make sure.

evgeni:
No, I don't think it would make it harder, as long as the standalone package will be higher version as the old provide (that will be removed)

PR #310, opened 2021-10-25
Update packages for Pulpcore 3.16

Odilhao:


Odilhao:
Depends on https://github.com/theforeman/tool_belt/pull/408

quba42:
I need to perform a pulp_deb release (but I can add that later)...

quba42:
pulp_deb 2.16 is now available (and should be added to the pulpcore 3.16 branch (at some point).
See: https://discourse.pulpproject.org/t/pulp-deb-2-16-0-has-been-released/209

Odilhao:
[test rpm]

packages/python-aiohttp-xmlrpc/python-aiohttp-xmlrpc.spec
```diff
@@ -5,17 +5,22 @@
...
 Name:           %{?scl_prefix}python-%{pypi_name}
-Version:        1.3.1
-Release:        2%{?dist}
+Version:        1.3.2
+Release:        1%{?dist}
 Summary:        aiohttp XML-RPC server handler and client
 
 License:        MIT
-URL:            https://github.com/mosquito/aiohttp-xmlrpc
+URL:            None
```
evgeni: this will explode ;)

package_manifest.yaml
```diff
@@ -164,6 +165,7 @@ pulpcore_packages:
     python-pulpcore-client: {}
     python-pycairo: {}
     python-pygobject: {}
+    python-pyopenssl: {}
```
evgeni: we have python-pyOpenSSL -- silly naming but that's what it is ;)

Odilhao: I added this while rebasing our branch rpm/3.16 using rpm/3.15 as base

evgeni: yeah, but we don't need it, we have it already. just different name

packages/python-jinja2/python-jinja2.spec
```diff
@@ -26,7 +26,7 @@ BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-setuptools
 %package -n     %{?scl_prefix}python%{python3_pkgversion}-%{srcname}
 Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{srcname}}
-Requires:       %{?scl_prefix}python%{python3_pkgversion}-markupsafe >= 0.23
+Requires:       %{?scl_prefix}python%{python3_pkgversion}-MarkupSafe >= 0.23
```
evgeni: the RPM doesn't provide that name, so you either need to lowercase it again, or fix the markupsafe rpm to add that provides

Odilhao:
[test rpm]

Odilhao:
[test rpm]

packages/python-aiohttp/python-aiohttp.spec
```diff
@@ -5,8 +5,8 @@
 %global pypi_name aiohttp
 
 Name:           %{?scl_prefix}python-%{pypi_name}
-Version:        3.7.4
-Release:        4%{?dist}
+Version:        3.7.4.post0
```
evgeni: please drop this bump, it's badly versioned and there are no user-relevant changes in this silly release (I hate python versioning so much)

packages/python-chardet/python-chardet.spec
```diff
@@ -5,8 +5,8 @@
 %global pypi_name chardet
 
 Name:           %{?scl_prefix}python-%{pypi_name}
-Version:        3.0.4
-Release:        3%{?dist}
+Version:        4.0.0
```
evgeni: aiohttp-3.7.4 requires chardet < 4.0, so this will break it.

post0 *does* allow 4.0, but your commit above didn't update the boundaries for some reason.

packages/python-aiohttp/python-aiohttp.spec
Odilhao: Ack

packages/python-chardet/python-chardet.spec
Odilhao: I'll drop this as well

Odilhao:
@evgeni it failed on repoclosure, but I think this was expected.

evgeni (approved):
repoclosure expected

PR #312, opened 2021-10-26
Obsolete the old Python 3.6 package for smooth upgrade - part X

evgeni:


ekohl (commented):
Am I reading it correctly that this is obsoleting on non-SCL (i.e., EL8). In case of bandersnatch (which has a `/usr/bin` entry) I can understand that but in the others? Aren't they co-installable?

evgeni:
bandersnatch and pkginfo have files in `/usr/bin` and thus conflict

django-read-only-fields depends on django and *that* is not co-installable due to files in `/usr/bin`

turtles all the way down

ekohl (approved):

PR #325, opened 2021-11-05
Remove python 3.6 package obsolete and exclude files in bin directory

simaishi:
Made the changes only to packages that caused conflicts upgrading galaxy, but excluded: ansible-builder, dynaconf, flake8, gunicorn, pulpcore and rq

### DIFF
--- a/packages/python-commonmark/python-commonmark.spec
+++ b/packages/python-commonmark/python-commonmark.spec
@@ -6,7 +6,7 @@
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        0.9.1
-Release:        3%{?dist}
+Release:        4%{?dist}
 Summary:        Python parser for the CommonMark Markdown spec
 
 License:        BSD-3-Clause
@@ -28,9 +28,6 @@ Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-future >= 0.14.0
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-setuptools
-%if 0%{?!scl:1}
-Obsoletes:      python3-%{pypi_name} < %{version}-%{release}
-%endif
 
 
 %description -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
@@ -63,12 +60,15 @@ set -ex
 %files -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 %license LICENSE
 %doc README.rst
-%{_bindir}/cmark
+%exclude %{_bindir}/cmark
 %{python3_sitelib}/%{pypi_name}
 %{python3_sitelib}/%{pypi_name}-%{version}-py%{python3_version}.egg-info
 
 
 %changelog
+* Fri Nov 05 2021 Satoe Imaishi - 0.9.1-4
+- Don't obsolete python 3.6 package and exclude files in bin
+
 * Wed Sep 29 2021 Evgeni Golov - 0.9.1-3
 - Obsolete the old Python 3.6 package for smooth upgrade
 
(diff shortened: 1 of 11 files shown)


simaishi:
@evgeni Not sure what the workflow is... what would trigger the rpm build (so I can test the galaxy upgrade with this change)?

Odilhao:
[test rpm]

packages/python-dynaconf/python-dynaconf.spec
```diff
@@ -61,12 +58,15 @@ set -ex
 %files -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 %license LICENSE
 %doc README.md
-%{_bindir}/dynaconf
+%exclude %{_bindir}/dynaconf
```
evgeni: I think we want `bin/dynaconf` to match what Pulp is using, and its "our" package

packages/python-gunicorn/python-gunicorn.spec
```diff
@@ -61,12 +58,15 @@ set -ex
 %files -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 %license LICENSE
 %doc docs/README.rst README.rst
-%{_bindir}/gunicorn
+%exclude %{_bindir}/gunicorn
```
evgeni: we need `bin/gunicorn`, otherwise app wont start.

packages/python-rq/python-rq.spec
```diff
@@ -64,14 +61,17 @@ set -ex
 %files -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 %license LICENSE
 %doc README.md
-%{_bindir}/rq
-%{_bindir}/rqinfo
-%{_bindir}/rqworker
+%exclude %{_bindir}/rq
+%exclude %{_bindir}/rqinfo
+%exclude %{_bindir}/rqworker
```
evgeni: we need these for the app.

evgeni:
Thanks Satoe!

Commented on the ones I think we *do* need the matching binaries from.

simaishi:
@evgeni thank you. Excluded dynaconf, gunicorn and rq from this change as well.

evgeni:
ok to test

evgeni:
For testing: when Jenkins is done building all of that, you'll see a long list of koji tasks show up in the "checks" section of the PR. each of them has a link to koji.katello.org, where you can then get the RPM, so you'll end up with a few clicks needed, and then you'll have the rpms to test your stuff with.

like this:

Screenshot from 2021-11-05 15-02-16

you obviously only need the EL8 builds ;)

packages/python-flake8/python-flake8.spec
```diff
@@ -68,12 +65,15 @@ set -ex
 %files -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 %license LICENSE
 %doc README.rst tests/fixtures/config_files/README.rst
-%{_bindir}/flake8
+%exclude %{_bindir}/flake8
```
ekohl: Looks like this is a dependency for `python-galaxy-importer`. Are you sure it only needs it as a library and not as an executable?

simaishi: Checking with the galaxy team...

ekohl: Looks like it is: https://github.com/ansible/galaxy-importer/blob/3315bb8aa76285aa5fb1623440f3f290fda4c89d/galaxy_importer/loaders/content.py#L129-L145

This may not be the only instance of it, just one that came to mind. Please check this for all other bins.

simaishi: Excluded `flake8` from this PR.

Spredzy:
@evgeni @Odilhao anything missing to proceed with that PR ?

evgeni (approved):
I'm fine with the changes. Will leave the merge to @Odilhao to not disturb the other work he's been doing.

PR #332, opened 2021-11-15
Update pulpcore to 3.16

Odilhao:


packages/python-drf-spectacular/python-drf-spectacular.spec
```diff
@@ -5,8 +5,8 @@
 %global pypi_name drf-spectacular
 
 Name:           %{?scl_prefix}python-%{pypi_name}
-Version:        0.17.3
-Release:        3%{?dist}
+Version:        0.21.0
```
evgeni: pulpcore has `Requires:       %{?scl_prefix}python%{python3_pkgversion}-drf-spectacular = 0.20.1`, how did 0.21.0 end up here?

evgeni (changes requested):

packages/python-drf-spectacular/python-drf-spectacular.spec
Odilhao: I just looked at the latest release and assumed that this was the right one while I was checking for reqs update on pulpcore. My bad

evgeni (approved):

PR #335, opened 2021-11-16
Update packages for 3.16 release

Odilhao:


Odilhao:
Repoclosure was expected to fail

evgeni (commented):
you could update the pulp-file entry in requirements.txt, but not blocking on this

Odilhao:
> you could update the pulp-file entry in requirements.txt, but not blocking on this

Done, updated the requirements.txt file.

PR #338, opened 2021-11-17
Use charset-normalizer instead of chardet

evgeni:


evgeni:
@Odilhao this should fix the issue you've been seeing

Odilhao (approved):

PR #339, opened 2021-11-18
Update python-galaxy-ng to 4.4.0

Odilhao:


evgeni:
Yyou need to manually remove the old tarball

Odilhao:
> Yyou need to manually remove the old tarball

Done, I forgot that obal get confused with galaxy-ng versioning

evgeni (approved):

PR #340, opened 2021-11-20
Don't use unversioned obsoletes

pcreech:
Unversioned obsoletes will cause issues with newer versions of some
toolings.

### DIFF
--- a/packages/python-dateutil/python-dateutil.spec
+++ b/packages/python-dateutil/python-dateutil.spec
@@ -7,7 +7,7 @@
 
 Name:           %{?scl_prefix}python-%{srcname}
 Version:        2.8.1
-Release:        4%{?dist}
+Release:        5%{?dist}
 Summary:        Extensions to the standard Python datetime module
 
 License:        Dual License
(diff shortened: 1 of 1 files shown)


packages/python-dateutil/python-dateutil.spec
```diff
@@ -29,7 +29,7 @@ BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-six >= 1.5
 Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{srcname}}
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-six >= 1.5
-Obsoletes:      %{?scl_prefix}python36-dateutil
+Obsoletes:      %{?scl_prefix}python36-dateutil < %{version}-%{release}
```
pcreech: @evgeni My main curiosity is if this was done "on purpose".  I'm not sure yet if we can get downstream package tooling to accept this.  It currently rejects the non-versioned format.

pcreech:
Don't merge without confirmation that the unversioned obsoletes is not intentional.

Odilhao (approved):

packages/python-dateutil/python-dateutil.spec
evgeni: This comes from the original epel dateutil package and can be dropped.

pcreech: switched up to drop the obsolete all together

PR #341, opened 2021-11-22
Update libcomps to 0.1.18

evgeni:


packages/libcomps/libcomps.spec
```diff
@@ -68,36 +61,16 @@ BuildRequires:  doxygen
 %description doc
 Documentation files for libcomps library.
 
-%package -n {?scl_prefix}python-%{name}-doc
+%package -n %{?scl_prefix}python-%{name}-doc
```
evgeni, line 64: We don't build docs, so we never noticed this, but… :)

evgeni:
This is needed to make updates to 3.15 work on EL 8.5, which ships with a newer libcomps than we did.

Odilhao (approved):

PR #343, opened 2021-11-24
Split devel files into own package

evgeni:


evgeni:
we only need the `cairo` part for Katello, but the devel files pull in additional dependencies (like `freetype-devel`) which we don't need and only add cruft to the system

Odilhao (approved):

PR #345, opened 2021-11-25
Remove 'obsolete' that was added previously for python 3.6 package

Odilhao:
and remove files in /usr/bin

(cherry picked from commit 71a74dd1d1ef4b6417562323a946ce6ac6c88da4)

### DIFF
--- a/packages/python-sqlparse/python-sqlparse.spec
+++ b/packages/python-sqlparse/python-sqlparse.spec
@@ -6,7 +6,7 @@
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        0.4.2
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        A non-validating SQL parser
 
 License:        BSD-3-Clause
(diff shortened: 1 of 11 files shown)


packages/python-sqlparse/python-sqlparse.spec
```diff
@@ -70,6 +67,9 @@ set -ex
 * Tue Nov 09 2021 Odilon Sousa <osousa@redhat.com> - 0.4.2-1
```
evgeni, line 70: I think *this* one needs to be bumped to 0.4.2-2, as it's actually been updated in 3.16, the others need just be tagged

packages/python-sqlparse/python-sqlparse.spec
```diff
@@ -6,7 +6,7 @@
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        0.4.2
-Release:        1%{?dist}
+Release:        2%{?dist}
```
evgeni, line 9: please add a changelog entry for that one :)

Odilhao: Done, my bad

evgeni (approved):
ACK

sqlparse will need a build, everything else needs to be tagged from 3.15

if you tag first, and then merge this PR, jenkins will do the right thing and only build sqlparse :)

Odilhao:
Done with all untags, merging this now

PR #349, opened 2021-12-03
Relase Galaxy-ng 4.2.7

Odilhao:

PR #351, opened 2021-12-08
Update python-pulp-ansible version requirement in galaxy_ng

simaishi:
pulp_ansible was updated to 0.5.10, along with galaxy_ng 4.2.7 (#349), but the version constraint didn't get updated in galaxy_ng, causing the install to fail.


Odilhao:
ok to test

evgeni (approved):

PR #371, opened 2022-01-25
bump epoch for gobject

evgeni:


packages/python-pygobject/python-pygobject.spec
```diff
@@ -70,5 +71,8 @@ set -ex
 
 
 %changelog
+* Tue Jan 25 2022 Evgeni Golov - 1:3.40.1-2
+- Bump epoch, there was a 3.42 version in the repo and users might have installed that.
```
ekohl, line 75: "the repo" is our repo?

Edit: from reading the other issue I do understand this, but it's a bit ambiguous. Especially when you consider that this will also end up in downstream.

evgeni: Yes.

ekohl:
This is my unfamiliarity with epochs so I do wonder: do dependency specifications need to include the epoch as well?

evgeni:
> This is my unfamiliarity with epochs so I do wonder: do dependency specifications need to include the epoch as well?

of fuck, yes.

Odilhao (approved):

PR #373, opened 2022-01-27
Update pulpcore to 3.14.11

Odilhao:


evgeni (changes requested):
you need to bump aiohttp-xmlrpc too: https://github.com/pulp/pulp_rpm/blob/3.14.11/requirements.txt#L9

the bump of pulp_container now makes the version greater than in 3.15 (which is fine, katello doesn't care)
but if you ever need to also bump it in 3.15, you can't just tag due to different python versions, this is why the previous build had a `0.1` as revision.

packages/python-pulp-rpm/python-pulp-rpm.spec
```diff
@@ -35,7 +35,7 @@ Requires:       python%{python3_pkgversion}-libcomps >= 0.1.15
 Conflicts:      python%{python3_pkgversion}-libcomps >= 0.2
 Requires:       python%{python3_pkgversion}-productmd >= 1.33
 Requires:       python%{python3_pkgversion}-pulpcore < 3.15
-Requires:       python%{python3_pkgversion}-pulpcore >= 3.14.5
+Requires:       python%{python3_pkgversion}-pulpcore >= 3.14.11
```
evgeni, line 38: can you also update the `aiohttp-xmlrpc` Requirement?

packages/python-aiohttp-xmlrpc/python-aiohttp-xmlrpc.spec
```diff
@@ -43,5 +43,8 @@ rm -rf %{pypi_name}.egg-info
 %{python3_sitelib}/aiohttp_xmlrpc-%{version}-py%{python3_version}.egg-info
 
 %changelog
+* Fri Jan 28 2022 Odilon Sousa <osousa@redhat.com> - 1.3.2-1
```
evgeni: this version already exists in 3.16, built against a different python, so you'd need to do `1.3.2-0.1`

packages/python-pulp-rpm/python-pulp-rpm.spec
Odilhao: Done, added on conflict as well

evgeni:
```
DEBUG util.py:600:  Error: Error downloading packages:
DEBUG util.py:600:    Status code: 404 for http://mirror.centos.org/centos/8/BaseOS/x86_64/os/Packages/binutils-2.30-108.el8_5.1.x86_64.rpm (IP: 198.15.72.18)
DEBUG util.py:753:  Child return code was: 1

```

@pcreech I thought we use local copies for all of that?

Odilhao:
[test rpm]

Odilhao:
\o/ @evgeni @pcreech

pcreech (approved):

pcreech:
@evgeni I approved this, as it appears the blocking changes you requested have been made, and it was time sensitive

PR #379, opened 2022-02-07
Updates patch for aiohttp to be compatible with 3.8.1

dkliban:


dkliban:
This is a very stale PR.  I am closing it.

PR #390, opened 2022-02-17
adjust galaxy-importer and ansible-lint dependencies for ansible-core in EL8

evgeni:
this way we don't depend on the exact package name, which can vary

### DIFF
--- a/packages/ansible-lint/ansible-lint.spec
+++ b/packages/ansible-lint/ansible-lint.spec
@@ -3,7 +3,7 @@
 
 Name:           %{pypi_name}
 Version:        5.0.8
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Checks playbooks for practices and behaviour that could potentially be improved
 
 License:        MIT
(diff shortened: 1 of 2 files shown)


zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

packages/ansible-lint/ansible-lint.spec
```diff
@@ -17,7 +17,7 @@ BuildRequires:  python%{python3_pkgversion}-setuptools-scm >= 3.5.0
 BuildRequires:  python%{python3_pkgversion}-setuptools_scm_git_archive
 
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
-Requires:       ansible >= 2.9
+Requires:       (ansible >= 2.9 or ansible-core)
```
ekohl: This is EL8 only, right? Should this be behind a conditional?

evgeni: No, we only build this package on EL8

zjhuntin (approved):

PR #391, opened 2022-02-17
Add python-smmap for python-gitdb

pcreech:


Odilhao:
Do we need to add this package to comps?

Odilhao (approved):

pcreech:
> Do we need to add this package to comps?

Added

PR #395, opened 2022-02-22
Add scl_prefix to soupsieve

Odilhao:


evgeni (dismissed):

packages/python-soupsieve/python-soupsieve.spec
```diff
@@ -1,46 +1,63 @@
+%{?scl:%scl_package python-%{pypi_name}}
+%{!?scl:%global pkg_name %{name}}
+
 # Created by pyp2rpm-3.3.7
 %global pypi_name soupsieve
 
 Name:           python-%{pypi_name}
```
evgeni: this needs to be 
```suggestion
Name:           %{?scl_prefix}python-%{pypi_name}
```

Odilhao: Done

evgeni (approved):

PR #398, opened 2022-02-22
Move from Vault to Centos8-Stream

Odilhao:
I'm opening this PR to move from Vault to Centos8Stream for repoclosure, the past week we had some timeouts with vault, it's not reliable and will only get older with time. 

el8-configmanagement-ansible was removed because the package `centos-release-ansible-29` is now provided by Extras.


zjhuntin:
I can also attest that this has been timing out frequently, and would be nice if it'd be more reliable.

zjhuntin (approved):

PR #399, opened 2022-02-23
Clean package_manifest from el8-configmanagement-ansible

Odilhao:


package_manifest.yaml
```diff
@@ -35,7 +35,6 @@ repoclosures:
           - el8-baseos
           - el8-appstream
           - el8-powertools
-          - el8-configmanagement-ansible
```
evgeni, line 38: there is another one in line 66

evgeni (changes requested):

package_manifest.yaml
Odilhao: Done

evgeni (approved):

PR #403, opened 2022-02-23
Fix Requires for pulp-rpm 3.17.3

Odilhao:


packages/python-pulp-rpm/python-pulp-rpm.spec
```diff
@@ -88,6 +88,9 @@ set -ex
 
 
 %changelog
+* Wed Feb 23 2022 Odilon Sousa <osousa@redhat.com> - 3.17.3-2
+- Fix Requires for pulpcore >= 3.16.0
+
 * Tue Feb 08 2022 Odilon Sousa <osousa@redhat.com> - 3.17.3-1
```
pcreech, line 94: 3.17.3-1 was already built for rpm/3.16:

https://github.com/theforeman/pulpcore-packaging/commit/7345480574182cb345f2bbf0c58b1d555e2e7af0

Instead of this PR, we may want to pick that commit forward and fix any discrepancies of updated requires.

Odilhao: 3.17.3-1 was also built for rpm/3.17, the problem that I see with picking that commit is that we need to change the Requires to < Puplcore 3.20 as well, and on rpm/3.16 it's < 3.18,  we will probably need to release a 3.17.3-2 anyway

PR #407, opened 2022-03-02
Proof of concept: Keep only python38-createrepo_c

hroncok:


ekohl:
Context for others: https://lists.centos.org/pipermail/centos-devel/2022-March/120250.html

I do think we'll need to keep building the libs for EL7 so it'll need to be a bit more complicated, but let's see what the build process shows
ok to test

packages/createrepo_c/createrepo_c.spec
```diff
@@ -195,63 +141,23 @@ pushd build-py3
...
-
-%if %{with python36}
-%files -n python3-%{name}
-/usr/lib64/python3.6/site-packages/%{name}/
-/usr/lib64/python3.6/site-packages/%{name}-%{version}-py*.egg-info
-%endif
+# Exclude everything shipped by the RHEL package
+%exclude %{_mandir}
+%exclude %{bash_completion}
+%exclude %{_libdir}
```
hroncok: Heh, this will likely exclude the Python bits as well, will amend.

ekohl:
Our CI requires a change in release and accompanying changelog entry. If you add those, our build system should run and we can see the results.

ekohl:
ok to test

ekohl:
@evgeni I'd like your thoughts on this.

evgeni:
We need the libs part on EL7 tho, as that comes with an ancient createrepo_c.
And we might get ahead of RHEL at some point (again) where we need a createrepo_c 0.18, which RHEL doesn't have yet.

But in general, yes, this is how we *should* do it ;)

evgeni:
Probably obsolete, now that pulp_rpm needs a newer createrepo than available in CentOS 8 anyways:

https://github.com/pulp/pulp_rpm/blob/10a67777685c6657263e30a6ca5550ba7cedcc48/requirements.txt#L1

Odilhao:
Closing this PR for now, we will soon rebuild everything with python 3.12.

PR #419, opened 2022-03-18
Update python-pulpcore to 3.16.5

ianballou:
This adds a required fix for a migration issue so Katello 4.4 can be released.


packages/python-pulpcore/python-pulpcore.spec
```diff
@@ -173,6 +173,9 @@ done
 
 
 %changelog
+* Fri Mar 18 2022 Ian Ballou <ianballou67@gmail.com> - 3.16.5-1
+- Release python-pulpcore 3.16.3
```
evgeni: Lies ;)

ianballou: whoops!

ianballou:
I ran all the Katello Pulpcore tests with this new version of Pulpcore, so we should be set.

jlsherrill (approved):

PR #425, opened 2022-03-28
Add provides and SCL wrappers for pulp-cli

evgeni:


packages/python-pulp-cli/python-pulp-cli.spec
```diff
@@ -63,17 +65,26 @@ set -ex
 %py3_build
 %{?scl:EOF}
 
+%if 0%{?scl:1}
+printf '#!/bin/bash\n%{?scl:source scl_source enable tfm-pulpcore \n}exec pulp "$@"\n' > pulp-cli-wrapper
```
ekohl, line 69: This `%{?scl:...}` looks a bit redundant given that it's already inside the SCL-only block, but it doesn't hurt either.

ekohl (approved):

packages/python-pulp-cli/python-pulp-cli.spec
evgeni: It is, it's copied from another spec and I was veeeeery lazy :)

PR #436, opened 2022-04-07
Update pulpcore-selinux to 1.3.0

ehelms:


evgeni (approved):
I guess this should be also picked into 3.17?

PR #438, opened 2022-04-07
[3.18] Build python-pytz against Python 3.9 on el8/9

dsavineau:
el7 isn't a target anymore so we don't need the SCL configuration.

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

### DIFF
--- a/packages/python-pytz/python-pytz.spec
+++ b/packages/python-pytz/python-pytz.spec
@@ -1,12 +1,9 @@
-%{?scl:%scl_package python-%{pypi_name}}
-%{!?scl:%global pkg_name %{name}}
-
 # Created by pyp2rpm-3.3.3
 %global pypi_name pytz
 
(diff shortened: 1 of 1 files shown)


zjhuntin:
[test rpm]

evgeni:
[test rpm]

evgeni (changes requested):
The diff looks sane if we say "we don't need SCLs anymore". However, in reality, we still have other branches that *do* need EL7/SCL to work. Dropping the SCL macro usage here means that we cannot easily backport changes done in the `3.18` branch to `3.17` and older.

If we don't drop them just yet, the EL8/EL9 builds will remain SCL-less (as the macros are undef there) but we'll be able to share changes with older branches, which I think is a great benefit.

So this PR should be really just a changelog entry and an increase of `%release`.

Spredzy:
This PR can be closed @Odilhao as we have been doing the work in other PRs

PR #442, opened 2022-04-14
mark Django 2.2.21+ not supported for pulp-container

evgeni:


evgeni:
this is due to the fact that epel8 now has a newer django, and our pipeline fail with that. I am not even sure this will help, but let's at least document it properly in the package :D

ekohl (approved):
Is it possible to specify a narrower range using rich dependencies?

evgeni:
> Is it possible to specify a narrower range using rich dependencies?

Maybe… I didn't try so far.

evgeni:
> I am not even sure this will help, but let's at least document it properly in the package :D

it did help, our pipelines are passing again, yay

PR #447, opened 2022-04-19
Update packages in rpms/3.18

Spredzy:


packages/python-certifi/python-certifi.spec
```diff
@@ -5,19 +5,17 @@
...
-Version:        2020.6.20
-Release:        2%{?dist}
+Version:        2021.10.8
+Release:        1%{?dist}
 Summary:        Python package for providing Mozilla's CA Bundle
 
 License:        MPL-2.0
 URL:            https://certifiio.readthedocs.io/en/latest/
 Source0:        https://files.pythonhosted.org/packages/source/c/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
-Patch0:         certifi-2020.6.20-use-system-cert.patch
```
evgeni: please don't drop this patch (or the `ca-certificates` requires), we want to use the system certs, not the bundled ones.

(in reality, you don't need to bump certifi because of this, as its content never changes for us)

packages/python-django-guid/python-django-guid.spec
```diff
@@ -5,11 +5,11 @@
 %global pypi_name django-guid
 
 Name:           %{?scl_prefix}python-%{pypi_name}
-Version:        3.2.1
+Version:        3.2.2
 Release:        1%{?dist}
 Summary:        Middleware that enables single request-response cycle tracing by injecting a unique ID into project logs
 
-License:        BSD
+License:        None
```
evgeni: that's not a valid license ;)

packages/python-pulp-rpm-client/python-pulp-rpm-client.spec
```diff
@@ -5,12 +5,12 @@
...
 
 Name:           %{?scl_prefix}python-%{pypi_name}
-Version:        3.17.3
+Version:        3.17.4
 Release:        1%{?dist}
 Summary:        Pulp 3 API
 
 License:        GPLv2+
-URL:            https://pulpproject.org
+URL:
```
evgeni: this breaks the linter at the moment :(

packages/python-pulpcore-client/python-pulpcore-client.spec
```diff
@@ -5,12 +5,12 @@
...
 
 Name:           %{?scl_prefix}python-%{pypi_name}
-Version:        3.17.3
+Version:        3.18.3
 Release:        1%{?dist}
 Summary:        Pulp 3 API
 
 License:        GPLv2+
-URL:            https://pulpproject.org
+URL:
```
evgeni: same, this breaks the linter :(

Spredzy:
[test rpm]

PR #456, opened 2022-04-25
Rebuilding for 3.9 from t to z

Odilhao:


packages/python-yarl/python-yarl.spec
```diff
@@ -6,18 +6,15 @@
...
 
 License:        Apache 2
 URL:            https://github.com/aio-libs/yarl/
 Source0:        https://files.pythonhosted.org/packages/source/y/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
 
 BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-devel
-BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-idna >= 2.0
-BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-multidict >= 4.0
 BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-setuptools
-BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-typing-extensions >= 3.7.4
```
pcreech, line 20: I'm curious as to why we are dropping dependencies on what sounds like a simple rebuild

Odilhao: Me and @Spredzy started to clean the dependencies that are not required, to keep maintainability in the future

pcreech: good enough for me.

pcreech (approved):

PR #469, opened 2022-04-27
Rebuilding libcomps against 3.9

Odilhao:
I'm not really sure on how to build on EL9, it's failing at the make stage, maybe @evgeni and @pcreech know what I'm missing

### DIFF
--- a/packages/libcomps/libcomps.spec
+++ b/packages/libcomps/libcomps.spec
@@ -25,7 +25,7 @@
 
 Name:           libcomps
 Version:        0.1.18
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Comps XML file manipulation library
 
 License:        GPLv2+
@@ -127,11 +127,19 @@ mkdir build-doc
 set -ex
 
 %if %{with python3}
+%if 0%{?rhel} == 7
 pushd build-py3
   # explicitly set INCLUDE_DIR and LIBRARY when inside the SCL, otherwise it's not found
   %cmake ../libcomps/ -DPYTHON_DESIRED:STRING=3 -DPYTHON_EXECUTABLE=%{__python3} %{?scl:-DPYTHON_INCLUDE_DIR=/opt/rh/rh-python38/root/usr/include/python3.8/ -DPYTHON_LIBRARY=/opt/rh/rh-python38/root/lib64/libpython3.8.so}
   %make_build
 popd
+%else
+pushd build-py3
+  # Build for el9
+  %cmake ../libcomps/ -DPYTHON_DESIRED:STRING=3 -DPYTHON_EXECUTABLE=%{__python3}
+  %cmake_build
+popd
+%endif
 %endif
 
 %if %{with python36}
@@ -160,7 +168,11 @@ set -ex
 
 %if %{with python3}
 pushd build-py3
+%if 0%{?rhel} == 7
   %make_install
+%else
+  %cmake_install
+%endif
 popd
 %endif
 
@@ -171,7 +183,7 @@ popd
 %endif
 %{?scl:EOF}
 
-
+%if 0%{?rhel} != 9
 %check
 %{?scl:scl enable %{scl} - << \EOF}
 set -ex
@@ -181,6 +193,7 @@ set -ex
 pushd build-py3
   make test
   make pytest
+%endif
 popd
 %endif
 
@@ -224,6 +237,9 @@ popd
 %endif
 
 %changelog
+* Wed Apr 27 2022 Odilon Sousa <osousa@redhat.com> - 0.1.18-2
+- Rebuilding Against python 3.9
+
 * Mon Nov 22 2021 Evgeni Golov - 0.1.18-1
 - Release libcomps 0.1.18
 


packages/libcomps/libcomps.spec
```diff
@@ -102,19 +102,17 @@ Python3 bindings for libcomps library.
...
-# workaround for https://github.com/rpm-software-management/libcomps/pull/64
-sed -i 's/EXACT//' libcomps/src/python/src/CMakeLists.txt
 
 # it can't detect our special PYTHONPATH and uses the compiled-in from the SCL Python
-%if 0%{?scl:1}
-sed -i "/OUTPUT_VARIABLE PYTHON_INSTALL_DIR/ s#))#).replace('rh/rh-python38', 'theforeman/tfm-pulpcore'))#" libcomps/src/python/src/CMakeLists.txt
-%endif
+#%if 0%{?scl:1}
+#sed -i "/OUTPUT_VARIABLE PYTHON_INSTALL_DIR/ s#))#).replace('rh/rh-python39', 'theforeman/tfm-pulpcore'))#" libcomps/src/python/src/CMakeLists.txt
+#%endif
```
pcreech: Since we're not building against el7 anymore, we can probably just drop any scl-related items.

pcreech: or, maybe not yet.  I don't know if we'll need to backport libcomps version updates to older pulp releases....  ignore this for now.

Odilhao: yeah, that is why I'm keeping that, maybe we need to backport later

evgeni: Please keep as much as possible:

https://github.com/theforeman/pulpcore-packaging/pull/438#pullrequestreview-945273742

Odilhao: I tried to keep as in the review for #438, and it failed to build for both versions

packages/libcomps/libcomps.spec
```diff
@@ -128,16 +126,16 @@ set -ex
 
 %if %{with python3}
 pushd build-py3
-  # explicitly set INCLUDE_DIR and LIBRARY when inside the SCL, otherwise it's not found
-  %cmake ../libcomps/ -DPYTHON_DESIRED:STRING=3 -DPYTHON_EXECUTABLE=%{__python3} %{?scl:-DPYTHON_INCLUDE_DIR=/opt/rh/rh-python38/root/usr/include/python3.8/ -DPYTHON_LIBRARY=/opt/rh/rh-python38/root/lib64/libpython3.8.so}
+#  # explicitly set INCLUDE_DIR and LIBRARY when inside the SCL, otherwise it's not found
+  %cmake ../libcomps/ -DPYTHON_DESIRED:STRING=3 -DPYTHON_EXECUTABLE=/usr/bin/python3
   %make_build
```
pcreech, line 134: is there a reason we are using `%make_build` over `%cmake_build` ??

What I can tell is, for el9, `__cmake_builddir` is set and added to the commands after a check for `__cmake_in_source_build`

```
  %__cmake \\\
        %{!?__cmake_in_source_build:-S "%{_vpath_srcdir}"} \\\
        %{!?__cmake_in_source_build:-B "%{__cmake_builddir}"} \\\
        -DCMAKE_C_FLAGS_RELEASE:STRING="-DNDEBUG" \\\
        -DCMAKE_CXX_FLAGS_RELEASE:STRING="-DNDEBUG" \\\
        -DCMAKE_Fortran_FLAGS_RELEASE:STRING="-DNDEBUG" \\\
        -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \\\
        -DCMAKE_INSTALL_DO_STRIP:BOOL=OFF \\\
        -DCMAKE_INSTALL_PREFIX:PATH=%{_prefix} \\\
        -DINCLUDE_INSTALL_DIR:PATH=%{_includedir} \\\
        -DLIB_INSTALL_DIR:PATH=%{_libdir} \\\
        -DSYSCONF_INSTALL_DIR:PATH=%{_sysconfdir} \\\
        -DSHARE_INSTALL_PREFIX:PATH=%{_datadir} \\\

```

Utilizing `%cmake_build` over `%make_build` should take this into account.

Odilhao: using %cmake_build built it, and using cmake_install also solved the install issue

packages/libcomps/libcomps.spec
```diff
@@ -127,17 +127,25 @@ mkdir build-doc
 set -ex
 
 %if %{with python3}
+%if 0%{?rhel} == 7
 pushd build-py3
   # explicitly set INCLUDE_DIR and LIBRARY when inside the SCL, otherwise it's not found
-  %cmake ../libcomps/ -DPYTHON_DESIRED:STRING=3 -DPYTHON_EXECUTABLE=%{__python3} %{?scl:-DPYTHON_INCLUDE_DIR=/opt/rh/rh-python38/root/usr/include/python3.8/ -DPYTHON_LIBRARY=/opt/rh/rh-python38/root/lib64/libpython3.8.so}
+  %cmake ../libcomps/ -DPYTHON_DESIRED:STRING=3 -DPYTHON_EXECUTABLE=%{__python3} %{?scl:-DPYTHON_INCLUDE_DIR=/opt/rh/rh-python39/root/usr/include/python3.9/ -DPYTHON_LIBRARY=/opt/rh/rh-python39/root/lib64/libpython3.9.so}
```
evgeni: el7 will never go to python39.

packages/libcomps/libcomps.spec
```diff
@@ -15,17 +15,17 @@
 # Our EL8 buildroots default to Python 3.8, but we also need a 3.6 build of libcomps
 # to make dnf happy
 %if 0%{?rhel} == 8
-%bcond_without python36
+%bcond_without python39
 %else
-%bcond_with python36
+%bcond_with python39
 %endif
```
evgeni: this will break installs, you need a 3.6 build on el8 to not to break system tools that use 3.6 and need libcomps

Odilhao: So we will need to have a built and publish version of libcomps for Pulp 3.18? Should I add the python36 devel to our buildroot?

packages/libcomps/libcomps.spec
```diff
@@ -171,7 +183,7 @@ popd
...
 %{?scl:EOF}
 
-
+%if 0%{?rhel} == 7
```
evgeni: why this `if`?

Spredzy (approved):

packages/libcomps/libcomps.spec
evgeni: as per IRC discussion, I think this should be `!= 9` if the tests fail only there for now

evgeni (approved):

PR #473, opened 2022-04-28
Missing Cherry-picks for 3.18

Odilhao:


packages/python-django/python-django.spec
```diff
@@ -6,8 +6,8 @@
 %global srcname django
 
 Name:           %{?scl_prefix}python-%{srcname}
-Version:        3.2.12
-Release:        2%{?dist}
+Version:        3.2.13
+Release:        1%{?dist}
```
evgeni, line 10: this release already exists: http://koji.katello.org/koji/buildinfo?buildID=70459

evgeni (changes requested):

packages/python-django/python-django.spec
Odilhao: So here's the problem, If I cherry pick and later bump in the same PR, our rpmcheck will fail, let me cherry-pick the change, and after the merge I'll bump the release to build against 3.9

evgeni: Yeah, I guess we could merge this, which will skip the build of django, as it's already there, and then do another bump for the actual rebuild

evgeni (approved):

packages/python-django/python-django.spec
Odilhao: Ack, and the cherry-pick was not clean from 3.17, because we did a bump later for python39

PR #484, opened 2022-05-06
Rebuild Packages with

Odilhao:
%{python_disable_dependency_generator} macro added to 3 packages

### DIFF
--- a/packages/python-aiohttp-socks/python-aiohttp-socks.spec
+++ b/packages/python-aiohttp-socks/python-aiohttp-socks.spec
@@ -1,12 +1,13 @@
 %{?scl:%scl_package python-%{pypi_name}}
 %{!?scl:%global pkg_name %{name}}
+%{?python_disable_dependency_generator}
 
 # Created by pyp2rpm-3.3.3
 %global pypi_name aiohttp-socks
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        0.7.1
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Proxy connector for aiohttp
 
 License:        Apache 2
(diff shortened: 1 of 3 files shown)


evgeni:
1. Why?
2. Does that work on el7?

Odilhao:
> 1. Why?

This is for EL9, those 3 packages right now have one auto requirement that was based on the requirements.txt, with a naming that doesn't march ours, it was auto added during build.

> 2. Does that work on el7?

No EL7 and EL8 will ignore this macro.

pcreech:
> 1. Why?

A little more detailed 'why':

On el9, python packages get auto-generated requires based on what is in the requirements.txt on the filesystem.  In most cases, this is 'fine', but in cases where requires specifies `extra` dependencies as well, this will generate a requires including this extra dependency.  

On the flip side, `extra` dependency `Provides:` don't get auto-generated.  We would need to generate this ourselves for packages that provide these extras by adding a line per extra to the spec.  In my research late last week, I could not easily locate anywhere in `pyp2rpm` that gives us this information, meaning we would need some more hands-on research to locate these packages.  The most effective stop-gap (in my opinion) is to turn off the auto-generated `Requires:` from these packages instead.  

Example requires output that was breaking el9:

```
python3.9dist(tablib) >= 3
python3.9dist(tablib[html]) >= 3
python3.9dist(tablib[ods]) >= 3
python3.9dist(tablib[xls]) >= 3
python3.9dist(tablib[xlsx]) >= 3
python3.9dist(tablib[yaml]) >= 3
```

evgeni:
Right, so I think this solves two issues:
1. different spelling (`django-rest-framework` vs `djangorestframework` is a very common example), which IMHO is better fixed by adding a `Provides` to the "misspelled" package
2. `extra` requires, which don't belong into *Requires* at all, which I agree can only be really fixed with this change.

read: ACK ;-)

PR #485, opened 2022-05-09
Adding a dependency for python-packaging at python-django-lifecycle

Odilhao:
Adding this dependency to unlock the installation more context here

https://github.com/pulp/pulp_installer/runs/6352353882?check_suite_focus=true#step:13:825

### DIFF
--- a/packages/python-django-lifecycle/python-django-lifecycle.spec
+++ b/packages/python-django-lifecycle/python-django-lifecycle.spec
@@ -6,7 +6,7 @@
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        0.9.6
-Release:        1%{?dist}
+Release:        2%{?dist}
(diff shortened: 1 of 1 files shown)


packages/python-django-lifecycle/python-django-lifecycle.spec
```diff
@@ -27,6 +27,7 @@ Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-django >= 2.0
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-urlman >= 1.2.0
+Requires:       %{?scl_prefix}python%{python3_pkgversion}-packaging >= 21.0
```
fao89, line 30: 👍🏽

fao89 (approved):
Thank you!

PR #499, opened 2022-05-12
Drop pin on %{_selinux_policy_version} for greater flexibility

Spredzy:
This leads to this package being only installable on equal or newer version of the system it has been built on.
By dropping that we allow for this package to be installed on older versions (LTS, EUS, ...)


Spredzy:
cc @Odilhao

evgeni:
Who is going to ensure it actually works with these older versions it was not built against?

ekohl:
Since this can't be guaranteed to work, I'm going to close this.

PR #510, opened 2022-05-24
Bump release for python-naya and createrepo_c for 3.18

Odilhao:
Also cherry-picking pulp-rpm


pcreech:
repoclosure failures are unexpected, but given current amount of rebuilds needed, we can evaluate after the build work is done.

PR #515, opened 2022-05-30
Changing gobject requirement for EL9 on pulp-ostree

Odilhao:


evgeni (approved):
lets try this

evgeni:
repoclosure only complains about the *existing* build, so let's see

PR #520, opened 2022-06-17
ostree: Obsolete the old Python 3.8 package for smooth upgrade

evgeni:


evgeni:
that should be the remaining bit for a smooth py39 upgrade on el8

cc @Odilhao

ehelms (approved):

PR #412, opened 2022-03-04
add python2-libcomps and libsolv0 to allow el7 to el8 upgrades

evgeni:
leapp depends on dnf, and dnf needs those older versions of packages to work on el7, but our "real" builds of libcomps and libsolv make it impossible to install the RPMs from extras (which leapp uses by default)

PR #541, opened 2022-08-10
forward-port python2-libcomps and libsolv0 to allow el7 to el8 upgrades

evgeni:
we had those changes in `rpm/3.16`, but not in 3.17+ as we initially planned not to have later Katello versions to be upgradeable from EL7 to EL8 (3.16 is used by 4.3, 4.4, 3.18 by 4.5)

this PR forward-ports the changes so that Katello 4.5 users can also upgrade their systems accordingly.

see #412 and #413 for the original PR and discussion

the packages are already tagged, this PR is for a) book-keeping and b) comps.xml


Odilhao:
I think we should exclude all packages for EL9

evgeni:
> I think we should exclude all packages for EL9

That'd be smart, wouldn't it?

Odilhao (approved):

PR #543, opened 2022-08-10
Force setuptools_scm usage for older setuptools for python-importlib-resources

Odilhao:
I was not able to do a clean cherry-pick, doing to we having too many branches and minor versions of the same package in different python version

:sob:


pcreech:
@Odilhao is this still needed?

PR #547, opened 2022-08-17
Bumping release to build against python 3.8 for pulp-cli

Odilhao:


Odilhao:
@m-bucher looks like we already have python-pulp-cli-0.14.0-2, it's the one built against python 39 for pulpcore 3.18, we will need another release bump

zjhuntin (approved):

PR #548, opened 2022-08-18
Bump python-debian to 0.1.44

quba42:
This contains new features that pulp_deb actually wants to use. :smile:


quba42:
Ideally I would also like to add this version to `rpm/3.16`. What is the best way of doing this? Cherry-pick all intervening version bumps? Or just create a separate version bump commit on the `rpm/3.16` branch?

Odilhao:
> Ideally I would also like to add this version to `rpm/3.16`. What is the best way of doing this? Cherry-pick all intervening version bumps? Or just create a separate version bump commit on the `rpm/3.16` branch?

3.16 is in a particular state right now, it uses python 3.8, we will need to rebuild with a upper release, since release 1 will be done against python 3.9. 

The easiest way is to create a new PR with the number 2 at the release.

Odilhao (approved):

PR #549, opened 2022-08-22
Bump python-debian to 0.1.44 (for rpm/3.16)

quba42:
Was already done for `rpm/3.18`: https://github.com/theforeman/pulpcore-packaging/pull/548

### DIFF
--- a/packages/python-debian/python-debian.spec
+++ b/packages/python-debian/python-debian.spec
@@ -6,8 +6,8 @@
 %global srcname debian
 
 Name:           %{?scl_prefix}python-%{srcname}
-Version:        0.1.42
-Release:        1%{?dist}
+Version:        0.1.44
+Release:        2%{?dist}
 Summary:        Debian package related modules
 
 License:        GPL-2+
@@ -28,6 +28,7 @@ BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-setuptools
 Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{srcname}}
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-chardet
+Requires:       zstd
 
 
 %description -n %{?scl_prefix}python%{python3_pkgversion}-%{srcname}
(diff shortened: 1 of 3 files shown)


packages/python-debian/python-debian.spec
```diff
@@ -67,6 +68,11 @@ set -ex
 
 
 %changelog
+* Mon Aug 22 2022 Quirin Pamp <pamp@atix.de> - 0.1.44-2
+- Update to 0.1.44
+- Add zstd dependency to support zstd compression for Ubuntu.
+- Use revision 2 to ensure build against python 3.8.
```
quba42, line 74: This is how I interpreted this comment: https://github.com/theforeman/pulpcore-packaging/pull/548#issuecomment-1219592745

Please double check if my interpretation makes sense/is correct!

packages/python-debian/python-debian.spec
```diff
@@ -67,6 +68,11 @@ set -ex
...
+
 * Wed Nov 03 2021 Odilon Sousa 0.1.42-1
```
quba42, line 76: I did not cherry-pick any of the intermediary revisions to this file from the `rpm/3.18` branch. Is that Ok, or should I have done so?

packages/python-debian/python-debian.spec
```diff
@@ -6,8 +6,8 @@
 %global srcname debian
 
 Name:           %{?scl_prefix}python-%{srcname}
-Version:        0.1.42
-Release:        1%{?dist}
+Version:        0.1.44
+Release:        2%{?dist}
```
quba42, line 10: Does it matter that this will result in a package version, that is higher than that built for the `rpm/3.18` branch? (`rpm/3.18` currently uses `0.1.44-1`)

Odilhao: We will need to bump again on 3.18 to avoid upgrade problems, having 2 python versions cause all sort of fun problems.

packages/python-debian/python-debian.spec
Odilhao: That's ok

packages/python-debian/python-debian.spec
quba42: Should I go ahead and open a PR for that? Should it bump the `Release` to 3?

Odilhao: Feel free to open that one, and we can merge this one after that

quba42: rpm/3.18 PR: https://github.com/theforeman/pulpcore-packaging/pull/551

Odilhao (approved):

PR #563, opened 2022-09-08
Bump pulp-cli release to rebuild for EL8

Odilhao:
We will need another bump for rpm/3.18


Odilhao:
Closing because we already have #547 open

PR #575, opened 2022-09-21
Use %{pypi_name} for %autosetup

goarsna:
We tried to package pulp-manifest using the spec file and the [pulp-manifest source code from github](https://github.com/pulp/pulp-manifest.git).

First we created a source distribution by `python3 setup.py sdist`. This creates a source distribution archive `pulp_manifest.tar.gz` which seems correct according to the `name` provided in `setup.py` and the `pypi_name` in the spec file.
Unfortunately at the `autosetup` step `pulp-manifest` is used hard coded. Therefore the build fails as the source distribution is written with an underscore and is therefore extracted to `pulp_manifest` as specified by the `pypi_name`.

IMO `pypi_name` should be used at the `autosetup` step.

### DIFF
--- a/packages/python-pulp-manifest/python-pulp-manifest.spec
+++ b/packages/python-pulp-manifest/python-pulp-manifest.spec
@@ -33,7 +33,7 @@ Requires:       %{?scl_prefix}python%{python3_pkgversion}-setuptools
 %prep
 %{?scl:scl enable %{scl} - << \EOF}
 set -ex
-%autosetup -n pulp-manifest-%{version}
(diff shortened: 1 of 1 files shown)


evgeni:
This would be correct if pulp_manifest would be uploaded to PyPI and we'd be building from an sdist source.
However https://pypi.org/project/pulp_manifest/ is 404 and we're building from the tarball GitHub generates for a tag (as seen in the Source0) line, and that uses the dash notation, not underscore.

goarsna:
Thanks for clarification. And in general thanks for your great work at pulpcore-packaging :)

PR #583, opened 2022-09-27
Changing dependencies to reflect the setup config for python >= 3.6

Odilhao:
https://opendev.org/opendev/bindep/src/tag/2.11.0/requirements.txt


Odilhao:
Repoclosure is expected

pcreech (approved):

PR #586, opened 2022-09-28
Update mock configs

ekohl:
This uses the mock configs from the host to reduce duplication. It also uses modularity in EL8 rather than using a flattened layout.

Additionally an EL9 mock config is created.

Besides the cherry pick of c70b84a16b6a7737a45ae21c291ba868f455632e this also updates the configs to point to 3.21.


mock/el8.cfg
```diff
@@ -1,67 +1,14 @@
...
-name=koji-non-modular
-baseurl=http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001-mod/non_modular/
-module_hotfixes=1
-
-[koji-python-38]
-name=koji-python-38
-baseurl=http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001-mod/python38:3.8/
-module_hotfixes=1
+config_opts['root'] = 'pulpcore-el8-x86_64'
+config_opts['chroot_setup_cmd'] += " python38-rpm-macros"
```
ehelms:
```suggestion
config_opts['chroot_setup_cmd'] += " python39-rpm-macros"
```

mock/el7.cfg
```diff
@@ -0,0 +1,15 @@
+include('centos-7-x86_64.cfg')
```
ehelms: I think this can be dropped for 3.21 as only EL8 and EL9 are build targets.

mock/el8.cfg
ekohl: I copied it from 3.18. Was 3.18 still Python 3.8 or was that already 3.9?

ehelms: Doh - looks like 3.18 was 3.9

ehelms (approved):

PR #593, opened 2022-10-04
Obsolete the old Python 3.8 package for smooth upgrade for pygobject

Odilhao:


packages/python-pygobject/python-pygobject.spec
```diff
@@ -31,7 +31,12 @@ Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{srcname}}
 Provides:       %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name} = %{version}
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-pycairo >= 1.16.0
-
+%if 0%{?!scl:1}
+Obsoletes:      python3-%{pypi_name} < %{version}-%{release}
```
sskracic, line 35: I don't think this will work as intended without %{epoch} which got bumped for the python39 build. Further, it should be %{srcname} and not {pypi_name}, as the former is fully lowercase, so it matches the RPM name.
```
Obsoletes: python3-%{srcname} < %{epoch}:%{version}-%{release}
...
Obsoletes:      python38-%{srcname} < %{epoch}:%{version}-%{release}
```

sskracic (changes requested):

zjhuntin (approved):

PR #596, opened 2022-11-01
Fixing typo to publish pulp_manifest for Pulpcore 3.18

Odilhao:


ekohl (commented):
Normally we also have the package directory and spec match the actual package name. That too has a dash instead of an underscore. Do you also plan to rename that so it's consistent again? And in that case you should also match it in `package_manifest.yaml`.

I'd prefer if everything was consistent with the conventions since some tooling may break if they don't.

Odilhao:
The package was branched like this on 3.14 and we didn't caught at the time, we already have it published as  python3-pulp_manifest for 3.14, I'll update the directory structure.

ekohl (approved):
Thanks! Let's not forget to also fix this in 3.21.

PR #613, opened 2023-01-26
Updating python-cryptography to 38.0.4

Odilhao:
This commit introduces the new rust dependecies, it vendors the rust crates and introduces the vendor_rust.py to vendor the crates.


Odilhao:
I want y'all to have fun reviewing this @pcreech @evgeni

ekohl (commented):
Any chance you can use the vendor packages from Fedora? I'm not sure if they're publicly published somewhere, but including 9 MB for every update in git will blow up the git repository.

Odilhao:
> Any chance you can use the vendor packages from Fedora? I'm not sure if they're publicly published somewhere, but including 9 MB for every update in git will blow up the git repository.

I don't see they being published, https://src.fedoraproject.org/rpms/python-cryptography/tree/rawhide and the version that Rawhide ships now it's older than then the version that we need, maybe we can do the vendor and host at https://yum.theforeman.org/ , but I'm not sure on how is our policy to upload vendor tarball there.

Odilhao:
I just created https://downloads.theforeman.org/vendor/cryptography-38.0.4-vendor.tar.gz to host the vendor content for python-cryptography. I'm still to create one script on tool-belt to sign the tarball for now I just added the md5.

zjhuntin (approved):
LGTM, thanks for figuring out that vendor issue, @Odilhao!

PR #631, opened 2023-02-15
Bump pulp-rpm to 3.19.0

ianballou:
I did this by hand so I'm not sure if the requirements are incorrect.


Odilhao:
We have some dependencies that need to be added:

https://github.com/pulp/pulp_rpm/compare/3.19.0...3.18.10#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R3

And could you add the 3.19.0 version to this file? https://github.com/theforeman/pulpcore-packaging/blob/rpm/3.22/automation/requirements.txt#L6

ianballou:
Sounds good. Also I'm adding in a pulp-file bump too, that needs to be updated as well.

ianballou:
Alrighty, I updated the requirements.

Odilhao (approved):

PR #632, opened 2023-02-16
Update the jsonschema requirement for Pulp-rpm 3.19.0

Odilhao:


packages/python-pulp-rpm/python-pulp-rpm.spec
```diff
@@ -42,7 +42,7 @@ Requires:       %{?scl_prefix}python%{python3_pkgversion}-createrepo_c >= 0.20.1
 Conflicts:      %{?scl_prefix}python%{python3_pkgversion}-createrepo_c >= 0.21
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-django-readonly-field >= 1.1.1
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-jsonschema >= 4.6
-Conflicts:      %{?scl_prefix}python%{python3_pkgversion}-jsonschema < 5.0
+Conflicts:      %{?scl_prefix}python%{python3_pkgversion}-jsonschema >= 5.0
```
ianballou, line 45: Is this not in conflict with the requirements here? https://github.com/pulp/pulp_rpm/blob/3.19/requirements.txt#L3

Odilhao: Fixed by changing to >= 5.1

ianballou: I'm still a bit confused, aren't the pulp-rpm requirements saying that jsonschema needs to be less than 5.0 ?

zjhuntin: The spec is currently indicating that the version of jsonschema required needs to be greater than or equal to 4.6, but will also conflict with any form of jsonschema that's less than 5.0 thus this won't work at all. This switch makes it, so it will conflict with any version of jsonschema that is greater than or equal to 5.0, but requires a version greater than or equal to 4.6.

zjhuntin (approved):

PR #648, opened 2023-05-03
Update python-pulp-deb to 2.20.2

quba42:


quba42:
Looks like foreman-packaging is already using the corresponding client for `rpm/develop` branch:
https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/katello/rubygem-pulp_deb_client

Question: If this is merged now, will this be part of the upcoming Katello 4.9 release? (That would be good)
https://community.theforeman.org/t/katello-4-9-branching-process/33357

ekohl:
@quba42 Katello 4.8 already uses the 3.22 repos, so it would also land there.

quba42:
@ekohl Then it might make sense to also update the client gem there: https://github.com/theforeman/foreman-packaging/tree/rpm/3.6/packages/katello/rubygem-pulp_deb_client

I believe this would not require a change to the Katello RPM, since it allows client gems up to the next Y release.

I do not know for certain if updating to this version of pulp_deb _requires_ the newer client, but there is reason to believe that it _might_ do so. (which is admittedly awkward, since this is a Z-release).

Should I go ahead and open a cherry-pick PR against foreman-packaging `rpm/3.6`?

quba42:
I went ahead and opened https://github.com/theforeman/foreman-packaging/pull/9336 to go with this.

Ideally both this and the foreman-packaging PR should be merged (or rather packaged) together.

evgeni (approved):

PR #652, opened 2023-06-22
Add configuration for Pulpcore 3.28 branching

Odilhao:


Odilhao:
I had these changes locally for the past 2 branches, I always forgot to push here, I'm changing the `Dockerfile` to use Centos8-stream, and also bumping the base python version to make sure that we only get the packages that are needed with python 3.9.

The `|| :` is to avoid the script failure with full PEP-517 packages.

pcreech:
@Odilhao Looks like this could use a rebase

PR #658, opened 2023-07-10
Update Pulpcore to 3.21.9 and Pulp-rpm to 3.18.17

Odilhao:


pcreech:
@Odilhao looks like this needs to be rebased

PR #676, opened 2023-07-19
Add packages for python-hatchiling build

Odilhao:


Odilhao:
`ptyprocess` and `pexpect` will need to be tagged, both packages got built for foreman.

packages/python-flit-core/python-flit-core.spec
```diff
@@ -30,7 +28,8 @@ Provides:       %{pypi_name} = %{version}
 %package -n     %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
 Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
-
+Requires:       %{?scl_prefix}python%{python3_pkgversion}-tomli
+Requires:       pyproject-rpm-macros
```
pcreech: Why is this a `Requires:` and not a `BuildRequires:` ?

Odilhao: Because we need this macros in every package that we build, we should add this to the BuildRoot in the future

pcreech: but... we only need that at build time yes?  Not runtime?

And why don't we just add it to the buildroot?

pcreech (approved):

PR #704, opened 2023-08-11
Add obsolete for django-current on python-django

Odilhao:


packages/python-django/python-django.spec
```diff
@@ -37,6 +37,14 @@ Obsoletes:      python3-%{srcname} < %{version}-%{release}
 Obsoletes:      python38-%{srcname} < %{version}-%{release}
 %endif
 
+%if 0%{?!scl:1}
```
evgeni, line 40: SCL, what SCL? :)

packages/python-django/python-django.spec
```diff
@@ -37,6 +37,14 @@ Obsoletes:      python3-%{srcname} < %{version}-%{release}
...
+## Obsoletes django-currentuser, the old package does not support Django 4
+Obsoletes:      python3-django-currentuser
+%endif
+%if 0%{?rhel} == 8
+Obsoletes:      python39-django-currentuser
```
evgeni, line 45: I'd prefer if such obsoletes would land in a `pulpcore-obsolete-packages` package, as Django as such doesn't really obsolete it (it breaks it, but it doesn't provide the same functionality).

Steal this album! https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec

zjhuntin:
@evgeni  https://github.com/theforeman/pulpcore-packaging/pull/705 @Odilhao is on PTO atm, but figure I can lend a hand :)

PR #705, opened 2023-08-15
Introduce pulpcore-obsolete-packages

zjhuntin:


packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
```diff
@@ -0,0 +1,32 @@
...
+BuildArch: noarch
+
+%if 0%{?!scl:1}
+Obsoletes:      python3-django-currentuser
+%endif
+%if 0%{?rhel} == 8
+Obsoletes:      python39-django-currentuser
+%endif
+
+Obsoletes: rubygem-foreman_column_view < 0.4.0-7
```
Odilhao: We don't need this here

packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
```diff
@@ -0,0 +1,32 @@
+Name: foreman-obsolete-packages
+Version: 1.0
+Release: 1%{?dist}
+License: MIT
+Summary: A package to obsolete retired packages
+URL: https://github.com/theforeman/pulpcore-packaging
+BuildArch: noarch
+
+%if 0%{?!scl:1}
```
Odilhao: let's remove the scl content, we don't need that anymore, the only if necessary is the one for `rhel8`

Odilhao:
You probably want to add later to python-pulpcore the dependency for this package.

We need to add this package to koji before merging.

Odilhao (approved):

evgeni:
> You probably want to add later to python-pulpcore the dependency for this package.

No need, dnf will pick this thanks to the obsoletes.

And now hush, go away! ;)

packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
```diff
@@ -0,0 +1,28 @@
...
+Version: 1.0
+Release: 1%{?dist}
+License: MIT
+Summary: A package to obsolete retired packages
+URL: https://github.com/theforeman/pulpcore-packaging
+BuildArch: noarch
+
+Obsoletes:      python3-django-currentuser
+%if 0%{?rhel} == 8
+Obsoletes:      python39-django-currentuser
```
evgeni, line 11: Can you update this to have an upper version bound, so if we ever want to introduce the package back, we can?

(Same for the el9 one)

evgeni (changes requested):

PR #706, opened 2023-08-28
Pin the obsolete version of django-currentuser

Odilhao:
This ensures that it's possible to add a new version of the package in the future

### DIFF
--- a/packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
+++ b/packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
@@ -1,14 +1,14 @@
 Name: pulpcore-obsolete-packages
 Version: 1.0
-Release: 1%{?dist}
+Release: 2%{?dist}
 License: MIT
 Summary: A package to obsolete retired packages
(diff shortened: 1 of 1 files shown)


Odilhao:
@evgeni just because you asked.

Thanks for asking :smile:

packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
```diff
@@ -1,14 +1,14 @@
...
 Version: 1.0
-Release: 1%{?dist}
+Release: 2%{?dist}
 License: MIT
 Summary: A package to obsolete retired packages
 URL: https://github.com/theforeman/pulpcore-packaging
 BuildArch: noarch
 
-Obsoletes:      python3-django-currentuser
+Obsoletes:      python3-django-currentuser <= 0.5.3
```
evgeni:
```suggestion
Obsoletes:      python3-django-currentuser < 0.5.3-6
```

The last build was 0.5.3-5, so we should make it like this :)

evgeni (approved):

PR #715, opened 2023-09-06
Update python-pulp-deb to 3.0.0 for rpm/3.28

quba42:
Built on top of #713 which should be merged and built first!

Will require a new clients gem. DO NOT MERGE until everything is ready and tested, and there is a Katello release that this can be synced up with!


quba42:
I believe there are two options for how to proceed:

1. Wait for a new pulpcore version repository > 3.28.
2. Add pulp_deb 3.0 to the 3.28 pulpcore repo, in this case, we need to:
   - Add new client gems to both rpm/develop, and rpm/3.8 in [foreman-packaging](https://github.com/theforeman/foreman-packaging/pull/9773) (contains Katello 4.10 which already uses pulpcore 3.28)
   - Make sure all new RPMs arrive in all the repositories at the same time, which must also coincide with a Katello 4.10.* release.

Irrespective of what Katello version first ends up using pulp_deb 3.0, we will need to test what this means for the Katello N-1 capsule/smart proxy promise.

Odilhao:
Katello 4.10 is already in the release process, 3.0 should land in develop 1st as usual and later we can pick into 4.10 branch at https://github.com/theforeman/foreman-packaging

quba42:
@Odilhao Unless we wait for a new pulpcore version repository I don't see how it is possible to add pulp_deb 3.0 to "develop" first. There is no pulpcore "develop" or "nightly" repository, only 3.28: http://yum.theforeman.org/pulpcore/

As soon as pulp_deb 3.0.0 lands in http://yum.theforeman.org/pulpcore/3.28/, everything that uses this repository and runs `dnf update` is broken, unless the new Katello RPM and the new client gem have also been added to http://yum.theforeman.org/katello/. Similarly, if we add the new client gem and Katello RPM to http://yum.theforeman.org/katello/nightly/ first, then this will be broken until pulp_deb 3.0.0 itself lands in http://yum.theforeman.org/pulpcore/3.28/.

=> Everything has to be added to all the repos simultaneously.

Satellite users might be protected from this effect by a version lock, but I don't believe Katello users are, or am I overlooking something?

Odilhao:
Maybe this is not the best place to discuss, maybe we should use discourse for this, tagging @ianballou just in case.

The workflow that I know for new features is go to `develop` in Foreman/Katello, if it breaks `develop` we should fix there before picking to one release in particular. 

From what I can see we should be able to update Katello `develop` to merge the bindings, after it gets merged on `develop` we merge it here on `pulpcore-packaging`, this will _unbreak_ develop. 

> Satellite users might be protected from this effect by a version lock, but I don't believe Katello users are, or am I overlooking something?

Katello 4.10 is not released yet, so we have time to break stuff IMHO, but I deffer that to @ianballou 

---

Edit: 3.28 is the version for 4.10, if we need a new release to make sure that 3.28 supports `pulp_deb` >= 3.0.0 it need to be cherrypick into 3.28. I know that pulp now release more often, but we don't have the bandwidth neither automation in place to update the rpms that often atm, that's why we picked 3.28 as the version for 4.10.

ianballou:
We can get this into Katello 3.28 as long as the testing for it in Katello starts now. I can assist with this and make a Katello PR.  @quba42 can you help with the client bindings update side of things? We'll need to:

- First, update the bindings for the gem itself
- Second, we'll need to bump up the bindings requirement in the katello spec file in foreman packaging.

I believe these two need to be different PRs.

At the same time I can make a Katello PR to bump up the versions from the Katello codebase itself and run the Pulpcore integration tests.

Now is the time to do this all because nightly is broken anyway on 4.10 not being released yet.

ianballou:
However, if I hit any blocking bugs during testing of pulp-deb 3.0.0, we'll probably need to delay bringing it into Katello.

quba42:
@ianballou I already created a foreman-packaging PR against develop: https://github.com/theforeman/foreman-packaging/pull/9773
I can create the analogous PR for foreman-packaging 3.8 as well?

The corresponding Katello PR is here: https://github.com/Katello/katello/pull/10719
This is against Katello master branch, so it would have to be cherry-picked to 4.10?
Also, I guess it needs an issue to close?

ianballou:
@quba42 inline responses:



> I can create the analogous PR for foreman-packaging 3.8 as well?

Yep!



> The corresponding Katello PR is here: [Katello/katello#10719](https://github.com/Katello/katello/pull/10719)
> This is against Katello master branch, so it would have to be cherry-picked to 4.10?

Yeah it'll need to be CP'd.  I'm looking to see if we have pulp-deb integration tests. If we do, they'll need to have their VCRs re-recorded against pulp-deb 3.0.0.  If I find any, I can record them and open a PR



> Also, I guess it needs an issue to close?

Yep!

quba42:
Also, @Odilhao can we start by merging the following, which do not need new clients:

https://github.com/theforeman/pulpcore-packaging/pull/714
https://github.com/theforeman/pulpcore-packaging/pull/713

But this PR is built on top of #713, so I need to rebase it once that is merged.

P.S.: @ianballou @Odilhao I think I can't get in touch with you in Foreman chat right now, since I am only using matrix. Trying to get the old hexchat configured now...

ianballou:
> P.S.: @ianballou @Odilhao I think I can't get in touch with you in Foreman chat right now, since I am only using matrix. Trying to get the old hexchat configured now...

I missed your messages since I was busy, but I opened up https://github.com/Katello/katello/pull/10723 so I could also update test fixtures.

ianballou:
This should be good to take out of draft.

PR #727, opened 2023-09-21
Update python-pulp-ansible to 0.18.1-1

ianballou:
To address the regression https://github.com/pulp/pulp_ansible/issues/1555


Odilhao:
Can we change the https://github.com/theforeman/pulpcore-packaging/blob/rpm/3.28/automation/requirements.txt#L7 line as well?

ianballou:
@Odilhao all set!

Odilhao (approved):

PR #732, opened 2023-10-02
Update django-import-export requirement 

Odilhao:


Odilhao:
Closing due to wrong ref

PR #737, opened 2023-10-13
Add pyproject-rpms-macros to Flit/Poetry to avoid adding to the build root

Odilhao:


Odilhao:
[test rpm]

Odilhao:
[test rpm]

Odilhao:
[test]

Odilhao:
[test rpm-copr]

Odilhao:
[test rpm-copr]

zjhuntin (approved):

packages/createrepo_c/createrepo_c.spec
```diff
@@ -1,53 +1,52 @@
...
-%if 0%{?rhel} && ( 0%{?rhel} <= 7 || 0%{?rhel} >= 9 )
+%if ( 0%{?rhel} && ( 0%{?rhel} <= 7 || 0%{?rhel} >= 9 ) ) || ( 0%{?fedora} && 0%{?fedora} >= 39 )
 %bcond_with drpm
 %else
 %bcond_without drpm
 %endif
 
-%if 0%{?rhel} || 0%{?fedora} < 29
+# Our EL8 buildroots default to Python 3.8, but let's also build 3.6, just to be safe
+%if 0%{?rhel} == 8
```
ekohl: If you check if Python 3 isn't 3.6 it also works where buildroots don't have another Python module defined:
```suggestion
%if 0%{?rhel} == 8 && 0%{python3_pkgversion} != 3
```

ekohl: Sorry, I should have tested this.
```suggestion
%if 0%{?rhel} == 8 && 0%{python3_pkgversion} != 03
```

Odilhao: Even with `&& 0%{python3_pkgversion} != 03` still fails in the verify version and release

ekohl: What is interesting is that the line number is off. It's trying to parse the old file (for comparison), and failing.

Odilhao: Yes, this is weird

ekohl: It's not weird: it's checking in `rpm/develop` and getting the version. Then it does the same in the PR. But that assumes you have working specs in `rpm/develop`.

For now I'd recommend a small PR that just makes sure `createrepo_c` works on plain EL8 and merge that, even if CI fails. Doesn't need to bump the release either and doesn't need to be built.

Then once it's merged, rebase this and I'd expect it to pass.

Odilhao: I just did a small PR, with the patch and it worked https://github.com/theforeman/pulpcore-packaging/pull/740

Odilhao:
Repoclosure is expect, we need to release createrepo_c

zjhuntin (approved):

PR #746, opened 2023-11-11
Mass rebuild packages with Python 3.11

Odilhao:


packages/ansible-lint/ansible-lint.spec
```diff
@@ -36,24 +40,31 @@ Requires:       python%{python3_pkgversion}-wcmatch >= 7.0
 %{summary}
 
 %prep
+set -ex
```
ekohl, line 43: Isn't `set -e` default behavior?

packages/ansible-lint/ansible-lint.spec
```diff
@@ -1,9 +1,12 @@
+%global python3_pkgversion 3.11
+%global __python3 /usr/bin/python3.11
+%{?python_disable_dependency_generator}
```
ekohl, line 3: Is there a specific reason to disable this?

packages/ansible-lint/ansible-lint.spec
```diff
@@ -1,9 +1,12 @@
+%global python3_pkgversion 3.11
+%global __python3 /usr/bin/python3.11
```
ekohl, line 2: Why are these needed? This should be set by `python3.11-rpm-macros`:
```console
# head -n 2 /usr/lib/rpm/macros.d/macros.python3.11
%__python3 /usr/bin/python3.11
%python3_pkgversion 3.11
```

packages/ansible-lint/ansible-lint.spec
```diff
@@ -13,8 +16,9 @@ BuildArch:      noarch
 
 BuildRequires:  python%{python3_pkgversion}-devel
 BuildRequires:  python%{python3_pkgversion}-setuptools
-BuildRequires:  python%{python3_pkgversion}-setuptools-scm >= 3.5.0
+BuildRequires:  python%{python3_pkgversion}-setuptools-scm >= 7.0.5
 BuildRequires:  python%{python3_pkgversion}-setuptools_scm_git_archive
+BuildRequires:  pyproject-rpm-macros
 
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
 # This should be
```
ekohl, line 24: We can now support rich dependencies so you can update it.

packages/createrepo_c/createrepo_c.spec
```diff
@@ -166,7 +169,7 @@ pushd build-py3
       -DWITH_LEGACY_HASHES=%{?with_legacy_hashes:ON}%{!?with_legacy_hashes:OFF} \
       -DWITH_ZSTD=%{?with_zstd:ON}%{!?with_zstd:OFF} \
       -DENABLE_DRPM=%{?with_drpm:ON}%{!?with_drpm:OFF} \
-      -DPYTHON_EXECUTABLE=/usr/bin/python3.9 -DPYTHON_LIBRARY=/usr/lib64/libpython3.9.so
+      -DPYTHON_EXECUTABLE=/usr/bin/python3.11 -DPYTHON_LIBRARY=/usr/lib64/libpython3.11.so
```
ekohl:
```suggestion
      -DPYTHON_EXECUTABLE=%{__python3} -DPYTHON_LIBRARY=/usr/lib64/libpython3.11.so
```

I also looked for a macro for the lib, but couldn't find a good one. This is also a weird one:
```
>>> import sys ;sys.version_info
sys.version_info(major=3, minor=11, micro=5, releaselevel='final', serial=0)
>>> sys.stdout.write('{0.major}.{0.minor}.'.format(sys.version_info))
3.11.5
>>> sys.stdout.write('{0.major}.{0.minor}'.format(sys.version_info))
3.114
```

This comes from:
```
%python3_version %(RPM_BUILD_ROOT= %{__python3} -Ic "import sys; sys.stdout.write('{0.major}.{0.minor}'.format(sys.version_info))")
```

I have no idea why it somehow adds the micro version.

packages/libsolv/libsolv.spec
```diff
@@ -232,10 +235,16 @@ set -ex
   -DPYTHON_EXECUTABLE=%{__python2}               \
 %if %{with python3_bindings}
   -DENABLE_PYTHON3=ON                            \
-  -DPYTHON3_EXECUTABLE=%{__python3}              \
+  -DPYTHON_EXECUTABLE=/usr/bin/python3.11        \
```
ekohl: Why not use `%{__python3}` to get the path?

packages/libsolv/libsolv.spec
```diff
@@ -232,10 +235,16 @@ set -ex
...
+  -DPYTHON_LIBRARY=/usr/lib64/libpython3.11.so.1.0 \
+  -DPYTHON_INCLUDE_DIR=/usr/include/python3.11   \
+  -DPython_ADDITIONAL_VERSIONS=3.11 \               
 %endif
 %else
   -DPYTHON_EXECUTABLE=%{__python3}               \
+  -DPYTHON_LIBRARY=/usr/lib64/libpython3.11.so.1.0 \
+  -DPYTHON_INCLUDE_DIR=/usr/include/python3.11   \
+  -DPython_ADDITIONAL_VERSIONS=3.11 \
```
ekohl, line 247: This one jumps out, are you sure this is supposed to be lowercase? And I don't see `ADDITIONAL` in the libsolv source code at all.

packages/libsolv/libsolv.spec
```diff
@@ -232,10 +235,16 @@ set -ex
   -DPYTHON_EXECUTABLE=%{__python2}               \
 %if %{with python3_bindings}
   -DENABLE_PYTHON3=ON                            \
-  -DPYTHON3_EXECUTABLE=%{__python3}              \
+  -DPYTHON_EXECUTABLE=/usr/bin/python3.11        \
+  -DPYTHON_LIBRARY=/usr/lib64/libpython3.11.so.1.0 \
```
ekohl, line 239: I don't see `PYTHON_LIBRARY` in the libsolv source code at all. Why is this needed?

packages/libsolv/libsolv.spec
```diff
@@ -232,10 +235,16 @@ set -ex
...
+  -DPYTHON_INCLUDE_DIR=/usr/include/python3.11   \
```
ekohl, line 240: I think this shouldn't be needed if you use the PYTHON3 variant: https://github.com/openSUSE/libsolv/blob/23cbed3219bd07b5c3fa1ed8a6f2fa6c478c0fdb/bindings/python3/CMakeLists.txt#L14 does this.

packages/libsolv/libsolv.spec
```diff
@@ -376,7 +385,7 @@ set -ex
 %files -n %{?scl_prefix}python%{python3_pkgversion}-%{libname}
 %{python3_sitearch}/_%{libname}.so
 %{python3_sitearch}/%{libname}.py
-%if 0%{?!scl:1}
```
ekohl, line 379: Should the whole spec file drop SCL support?

packages/ansible-lint/ansible-lint.spec
Odilhao: Yes, when we don't disable this on EL9 the dependency generation gets a little complicated since some of our packages don't generate the same metadata, it creates a repoclosure problem.

packages/ansible-lint/ansible-lint.spec
Odilhao: It's a `feature` of 3.11, the macros don't expand on EL9, only on EL8, on EL9 it defaults to python3 which is python3.9, me and @evgeni had a little ride last friday with this.

packages/ansible-lint/ansible-lint.spec
Odilhao: How would be the best to change this?

packages/libsolv/libsolv.spec
Odilhao: When I tried with python3, it built using python 3.9 on EL9, I'm really no sure whats going on there

packages/libsolv/libsolv.spec
Odilhao: If I don't add this, it builds with python 3.9 on EL9

packages/libsolv/libsolv.spec
Odilhao: Probably, I was going to look at scl after the release

Odilhao:
The packages got built here https://copr.fedorainfracloud.org/coprs/g/theforeman/pulpcore-nightly-staging/packages/

packages/ansible-lint/ansible-lint.spec
Odilhao: We have set -ex in all packages for pulpcore

packages/createrepo_c/createrepo_c.spec
Odilhao: Ok, scratch build using `%{__python3}` worked 

https://copr.fedorainfracloud.org/coprs/g/theforeman/pulpcore-nightly-staging-scratch-2cdd8425-0950-5979-bea5-035f5b6c9072/build/6629861/ 

Commit Added

packages/ansible-lint/ansible-lint.spec
ekohl: Oh, I only checked EL8. Isn't that a bug in RPM packaging? Looks like `python3.11-devel` provides `python3.11-rpm-macros` but doesn't include and actual RPM macros.

ekohl (changes requested):
We declare the module dependencies here:
https://github.com/theforeman/pulpcore-packaging/blob/9ac98f86751efcc12c09f13c21d1cecba71a5fc0/modulemd/modulemd-pulpcore-el8.yaml#L16-L21

At the least, we should drop that. But if we no longer need modularity for Python 3.11, does it still make sense to have a pulpcore module?

packages/ansible-lint/ansible-lint.spec
ekohl: I think the suggestion is in the file itself. @evgeni any thoughts?

packages/ansible-lint/ansible-lint.spec
Odilhao: It includes the macro, but it does not override the python3 macro on EL9 that is python3.9

Odilhao:
Modules dropped with 240f0b8ad997734c97b97f01190d0921ea6fc63d

packages/libsolv/libsolv.spec
Odilhao: Maybe this is a flag that is not used anymore, let me point fingers at @evgeni he probably knows

packages/libsolv/libsolv.spec
ekohl: You may not have set `PYTHON3_EXECUTABLE` so it was using `/usr/bin/python3` instead of `/usr/bin/python3.11`.

packages/libsolv/libsolv.spec
Odilhao: Changed to `%{__python3} `  and it worked now with e039215dce108e8f74d3453b5404a3186d747d79

packages/ansible-lint/ansible-lint.spec
Odilhao: pyproject-rpm-macros is a project we build for EL8, it's present on EL9 by default, but we need this for PEP-517, we build this internally

packages/libsolv/libsolv.spec
pcreech: Is it not enough to just switch -DPYTHON3_EXECUTABLE=/usr/bin/python3.11 ??  And does it still break even though we are resetting these macros at the beginning?

pcreech: and PYTHON_LIBRARY appears to be a function of CMAKE itself, helping specify which python library to use.  Probably useful with code generation with swig https://cmake.org/cmake/help/latest/module/FindPythonLibs.html

packages/ansible-lint/ansible-lint.spec
ekohl: I'm not talking about the BuildRequires, but about the comment (and the lines that follow).

Odilhao: Oh ok, we would need to work in all packages, because we use the same line for all packages.

packages/libsolv/libsolv.spec
Odilhao: As pointed by @pcreech it's a cmake flag https://cmake.org/cmake/help/latest/module/FindPythonLibs.html

Odilhao:
@ekohl can we merge this? I want to address the other pulp bits today, and later this week work on the obsoletes that we need for upgrades from python39 to 3.11

pcreech:
So, reviewing this pr, and the trade-offs of needing something functional to keep other trains moving smoothly, I'm going ahead and merging this PR.  It appears most everything should be mostly functional to a degree good enough to iterate on top of to fix issues that crop up.


Anything that is felt should have been addressed before merging this PR, let's identify them and create action items to follow up on asap after.

PR #750, opened 2023-11-15
Various python 3.11 cleanups

pcreech:


pcreech:
repoclosure failure on rpm-copr is because some latent pointers to koji.  Merging.  feel free to retro-ack or retro-nack

PR #755, opened 2023-11-16
Add egg again for python-3.11 builds on libcomps

Odilhao:


pcreech (approved):
It does appear the 3.11 package once again has appropriate provides and contains egg info

PR #790, opened 2023-11-22
Don't obsolete python3-pyyaml

pcreech:


packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
```diff
@@ -1,13 +1,12 @@
 Name: pulpcore-obsolete-packages
 Version: 1.0
-Release: 3%{?dist}
+Release: 4%{?dist}
 License: MIT
 Summary: A package to obsolete retired packages
 URL: https://github.com/theforeman/pulpcore-packaging
 BuildArch: noarch
```
Odilhao, line 8: Let's add `%if 0%{?rhel} == 9` to avoid problem with system python on EL8

Odilhao (approved):

packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
pcreech: Let's not... as we will probably hit the issue necessitating this on el9 as well.  python3-pyyaml is also provided by the system, and if we obsolete it, it may dissappear

PR #804, opened 2023-11-27
Update python-urllib3 to 1.26.18

Odilhao:


pcreech:
I'm still concerned by the what appears to be valid repoclosure issues, even though they don't immediately seem to stem from this PR.  do we have an explanation yet?

PR #808, opened 2023-11-29
Add {?dist} to python-json-stream

Odilhao:


packages/python-json-stream/python-json-stream.spec
```diff
@@ -8,7 +8,7 @@
 
 Name:           python-%{pypi_name}
 Version:        2.3.2
-Release:        2
+Release:        3{?dist}
```
pcreech: Missing % starting dist macro

pcreech (approved):

PR #812, opened 2023-12-08
Drop python-pulp_2to3_migration

ekohl:
The last supported Pulpcore version was 3.19 so this plugin and its dependencies can be cleaned up.


Odilhao:
Can we get the same packages added into [pulpcore-obsolete-packages](https://github.com/theforeman/pulpcore-packaging/blob/4624fb179f838e8087fadde94515234514b5618f/packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec) since it was present in 3.18, some old installation will probably have the python39 build.

ekohl:
Updated. Pulp 3.14 was the last release to ship it (https://yum.theforeman.org/pulpcore/3.14/el8/x86_64/) so I've special cased it. Not entirely sure this is correct, so please review carefully.

Odilhao:
We need to update the Version for obsoletes to build https://github.com/theforeman/pulpcore-packaging/blob/3ee22166b9a227c4bbf655b26b4007d667189ad8/packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec#L3

quba42:
What is missing for this cleanup to proceed?

ekohl:
Rebased

quba42 (approved):
I am not a person with merge access, but I like cleanup, and this is definitely no longer needed, so I am going to go ahead and leave an approving review. This all LGTM!

packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
```diff
@@ -1,13 +1,16 @@
...
+Version: 1.2.0
 Release: 1%{?dist}
 License: MIT
 Summary: A package to obsolete retired packages
 URL: https://github.com/theforeman/pulpcore-packaging
 BuildArch: noarch
 
 Obsoletes:      python3-django-currentuser < 0.5.3-6
+Obsoletes:      python3.11-pymongo
+Obsoletes:      python3.11-mongoengine
```
evgeni: those should be versioned, right? :)

packages/pulpcore-obsolete-packages/pulpcore-obsolete-packages.spec
```diff
@@ -1,13 +1,16 @@
...
 %if 0%{?rhel} == 8
+Obsoletes:      python3-pulp-2to3-migration < 0.12.0-3
```
evgeni, line 13: I was confused at first, but there were really no other names of this pkg:
```
[evgeni@web01 pulpcore]$ find . -name \*2to3\*noarch.rpm | xargs -n1 basename | sed 's/-0\..*//' | sort -u
python3-pulp-2to3-migration
```

evgeni: Technically, there is one el9 build:
```
[evgeni@web01 pulpcore]$ find . -name \*2to3\*el9\*noarch.rpm
./3.18/el9/x86_64/python3-pulp-2to3-migration-0.12.0-1.el9.noarch.rpm
```
but we never supported that release on el9 for Katello, so we don't need to care, IMHO

evgeni (approved):

Odilhao (approved):

PR #820, opened 2023-12-12
Remove overzealous obsoletes from packages

pcreech:
python-aiodns through python-certifi

### DIFF
--- a/packages/python-aiodns/python-aiodns.spec
+++ b/packages/python-aiodns/python-aiodns.spec
@@ -8,7 +8,7 @@
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        3.0.0
-Release:        5%{?dist}
+Release:        6%{?dist}
 Summary:        Simple DNS resolver for asyncio
 
 License:        MIT
@@ -28,10 +28,6 @@ BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-setuptools
 Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-pycares >= 4.0.0
-
-%if 0%{?rhel} == 8
-Obsoletes:      python39-%{pypi_name} < %{version}-%{release}
-%endif
 
 
 %description -n %{?scl_prefix}python%{python3_pkgversion}-%{pypi_name}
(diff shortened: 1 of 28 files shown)


packages/python-ansible-builder/python-ansible-builder.spec
```diff
@@ -76,6 +73,9 @@ set -ex
 
 
 %changelog
+* Tue Dec 12 2023 Patrick Creech <pcreech@redhat.com> - 1.0.1-7
+- Rollback overzealous obsoletes
```
evgeni: this one contains things in `_bindir` and *needs* the obsoletes.

packages/python-bindep/python-bindep.spec
```diff
@@ -74,6 +70,9 @@ set -ex
 
 
 %changelog
+* Tue Dec 12 2023 Patrick Creech <pcreech@redhat.com> - 2.11.0-5
+- Rollback overzealous obsoletes
```
evgeni: bindir!

packages/python-ansible-builder/python-ansible-builder.spec
pcreech: good catch!  I thought I skipped them all, guess not.  Will fix when I come back around to clean up

packages/python-bindep/python-bindep.spec
pcreech: See previous comment

evgeni (changes requested):
the packages that have things in `_bindir` and `_sbindir` should keep their obsoletes.

packages/python-bindep/python-bindep.spec
pcreech: er... "%exclude %{_bindir}/bindep"  - but... we're not including a bindir, we're excluding it?

pcreech: per irc convo, we're gonna keep the Obsoletes on anything with bindir for now, and can remove as we find need to.

packages/python-ansible-builder/python-ansible-builder.spec
pcreech: removed

packages/python-bindep/python-bindep.spec
pcreech: removed

Odilhao (approved):

Odilhao:
I reviewed all files, we should be good to merge this after the scratch pipeline, before the merge I'll clean the build on copr to avoid obsolete problems

PR #822, opened 2023-12-12
Remove overzealous obsoletes from packages

pcreech:
python-drf-spectacular through python-msrest

### DIFF
--- a/packages/python-drf-spectacular/python-drf-spectacular.spec
+++ b/packages/python-drf-spectacular/python-drf-spectacular.spec
@@ -9,7 +9,7 @@
 
 Name:           %{?scl_prefix}python-%{pypi_name}
 Version:        0.26.5
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Sane and flexible OpenAPI 3 schema generation for Django REST framework
 
 License:        BSD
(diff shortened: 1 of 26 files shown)


packages/python-enrich/python-enrich.spec
```diff
@@ -33,10 +33,6 @@ Requires:       %{?scl_prefix}python%{python3_pkgversion}-rich >= 9.5.1
 
 Obsoletes:      python3-%{pypi_name} < %{version}-%{release}
 
-%if 0%{?rhel} == 8
```
Odilhao: We should keep the obsolete here

packages/python-galaxy-importer/python-galaxy-importer.spec
```diff
@@ -56,9 +56,6 @@ Requires:       tar
 
 Obsoletes:      python3-%{pypi_name} < %{version}-%{release}
 
-%if 0%{?rhel} == 8
```
Odilhao: We should keep the obsolete here

packages/python-iniparse/python-iniparse.spec
```diff
@@ -44,9 +44,6 @@ BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-six
 BuildRequires:  %{?scl_prefix}python%{python3_pkgversion}-test
 Requires:       %{?scl_prefix}python%{python3_pkgversion}-six
 Obsoletes:      %{?scl_prefix}platform-python-%{modname} < %{version}-%{release}
-%if 0%{?rhel} == 8
```
Odilhao: We should keep the obsolete here

Odilhao (approved):
Partial approval, we need to keep the obsolete for some packages

PR #828, opened 2023-12-14
Remove obsolete from more packages on Pulp 3.39

Odilhao:


packages/python-charset-normalizer/python-charset-normalizer.spec
```diff
@@ -68,6 +64,9 @@ set -ex
 
 
 %changelog
+* Thu Dec 14 2023 Odilon Sousa <osousa@redhat.com> - 2.1.1-4
+- Dont obsolete python-charset-normalizer
```
evgeni, line 68: but it has something in bindir?

evgeni (approved):

PR #829, opened 2023-12-15
Obsolete python-rich for smooth upgrade

Odilhao:


evgeni (commented):
I guess we'll also have to add an obsolete in python39-enrich-1.2.6-5.el8.noarch, as otherwise that becomes unsatisfied

Odilhao:
> I guess we'll also have to add an obsolete in python39-enrich-1.2.6-5.el8.noarch, as otherwise that becomes unsatisfied

It's already there
https://github.com/theforeman/pulpcore-packaging/blob/5a63954bd478c2fc2912a655a7737be3e1c9f89b/packages/python-enrich/python-enrich.spec#L36

evgeni (approved):

PR #849, opened 2024-01-12
Release pulpcore-selinux 2.0.1

evgeni:


packages/pulpcore-selinux/pulpcore-selinux.spec
```diff
@@ -60,7 +61,7 @@ do
   done
 done
 systemctl daemon-reexec &>/dev/null || :
-/sbin/fixfiles -R python3-pulpcore restore || :
+/sbin/fixfiles -R python%{pulpcore_python}-pulpcore restore || :
```
evgeni, line 64: https://bugzilla.redhat.com/show_bug.cgi?id=2131798

Odilhao (approved):

packages/pulpcore-selinux/pulpcore-selinux.spec
Odilhao: Thanks, let me update that BZ with the right version

PR #863, opened 2024-01-23
Rebase libsolv and add epoch

pcreech:


packages/libsolv/libsolv.spec
```diff
@@ -1,60 +1,59 @@
-%global python3_pkgversion 3.11
```
Odilhao, line 1: Do we really need to drop the python 3.11 bindings?

pcreech: we probably want to _keep_ it even :D  Hence why I [wip] 'ed it

I did the initial copy over, but have not had a chance to update it to also build for python 3.11

packages/libsolv/libsolv.spec
```diff
@@ -376,64 +274,569 @@ set -ex
...
-/usr/lib64/python3.6/site-packages/_%{libname}.so
-/usr/lib64/python3.6/site-packages/%{libname}.py
-/usr/lib64/python3.6/site-packages/__pycache__/%{libname}.*
-%endif
 
 %changelog
-* Mon Nov 20 2023 Patrick Creech <pcreech@redhat.com> - 0.7.22-6
-- Obsolete python39-libsolv as well
+* Mon Jan 22 2024 Patrick Creech <pcreech@redhat.com> - 1:0.7.20-6
+- Rebase libsolv and add epoc
```
ekohl, line 286:
```suggestion
- Rebase libsolv and add epoch
```

packages/libsolv/libsolv.spec
```diff
@@ -1,60 +1,59 @@
...
+
+#global commitnum 2901
+#global commit 47fbaa2a0892866d30ec0e1b4c885532d0aca7b8
+#global shortcommit %%(c=%%{commit}; echo ${c:0:7})
+
+%define __cmake_switch(b:) %{expand:%%{?with_%{-b*}:ON}}%{expand:%%{!?with_%{-b*}:OFF}}
 
 Name:           lib%{libname}
-Version:        0.7.22
+Version:        0.7.20
```
evgeni: Why are we downgrading things?

evgeni: additionally: how are we going to handle EL9 which ships libsolv-0.7.24 today.

evgeni: and what about https://github.com/pulp/pulp_rpm/blob/main/requirements.txt#L7 (this has been true since pulp_rpm 3.17, which is pulpcore 3.16+)

pcreech: I have confirmed with pulp that while the requirements.txt for pip is 0.7.21, the functionality required should be back-ported into the OS provided libsolv.  We should be able to modify that requirements.txt via spec to loosen the requirement. (it's not exactly true in a purely upstream pip installed context, hence the need for spec modificaiton).

Ideally, we should probably _stop_ shipping libsolv on el9.  Revert back to using system libsolv, and only ship 3.11 bindings for it.  As far as "how", i'm not entirely sure.  I need to do a little research in this space.

evgeni: Yeah, if we can avoid building the lib and only do the bindings, that'd be awesome.
But the last time I looked at this, I ran away pretty quickly, as the bindings are essentially generated from the lib source and you can't build them w/o the lib (or, if you build both and only ship the bindings, who the hell guarantees that the lib from the OS and the bindings from the lib  you don't ship actually match up)

pcreech: Ok, so after some research and some help from fedora-devel matrix channel, we can build the python binding in a statically linked way.  I played around with this locally, and it also appears to be the way solv is distributed by pypi.  Therefore it won't override system libsolv, but won't use it either.  Meaning we can safely decouple from system libsolv, unless we have good reason to keep it coupled.

evgeni: Ah, so you build a pythonX-solv that contains a copy of libsolv but that's only used inside the Python X context? Fair

pcreech: Yup :D best way out of this mess I can figure out.

pcreech:
Still a little "WIP", but I think I have the strategy. 

We introduce "libsolv" with an epoch, but otherwise 1-1 with system libsolv.  This _should_ fix the el8 issue. (still need to fix it to only build on el8)

We also introduce a 'python-solv' package for both el8 and el9, which provides a python3.11-solv.  This *should* allow us to not be version locked with system libsolv, and keep us out of the game of dealing with an epoch on el9, and we can get out of this mess with libsolv when we drop el8.

packages/python-solv/python-solv.spec
```diff
@@ -0,0 +1,150 @@
...
+
+%check
+%ctest
+
+# Python smoke test (not tested in %%ctest):
+export PYTHONPATH=%{buildroot}%{python3_sitearch}
+export LD_LIBRARY_PATH=%{buildroot}%{_libdir}
+%python3 -c 'import solv'
+
+%files
```
evgeni: I'd argue the resulting binary package should be called `python3.11-solv`, not `python-solv`.

evgeni:
So I briefly tested the EL9 python-solv package, and besides the wrong name, it seems to have a working python solv stack, without being linked to the system libsolv. Yay.

Just to unfuck my brain, the plan is to:
* On EL8
  * Ship libsolv rebased on the version from EL8, but with an epoch to effectively force a downgrade from the version we currently have
  * Ship python3.11-solv built from the special sauce here
* On EL9
  * Not ship any of the libsolv things (EL9 is newer than what we have today) [TODO: This is not what this PR does right now]
  * Ship python3.11-solv built from the special sauce here

Question: Would a version like `0.7.22+really+0.7.20` for libsolv on EL8 be of any benefit (I hate epochs), as it would allow Base EL to ship a 0.7.z with z>22 that will naturally upgrade from ours, eliminating our need to keep uptodate with EL patches (as that is what we're buying here).

packages/python-solv/python-solv.spec
pcreech: +1

pcreech:
> Question: Would a version like 0.7.22+really+0.7.20 for libsolv on EL8 be of any benefit (I hate epochs), as it would allow Base EL to ship a 0.7.z with z>22 that will naturally upgrade from ours, eliminating our need to keep uptodate with EL patches (as that is what we're buying here).

Sadly, no.  There is not a way to "upgrade" out of a module, and we are in the module at this point.  I'm not a fan of epochs either, but this is kinda the reason they exist in RPM land, and translate better.  We will never "upgrade" out of this state sadly.

--------------------------

Currently fleshing out the small updates, should be g2g soon

pcreech:
4051d184b0aab14d7d8987930de3c7ae008b565d will need to be removed once we are ready to merge.  This commit is only to bypass the version/release check that keeps failing due to rebasing on a new version/release with epoc.  It really doesn't like the release for some reason.

Odilhao:
[rpm-copr test]

evgeni (changes requested):
python3-solv is still in the el9 comos, but should not?

evgeni:
[test rpm-copr] as I have no idea why it failed and the logs are gone

evgeni:
el9 comps still contain `ruby-solv` and `python3-solv`.

evgeni:
aha, building `libsolv` on EL8 fails: `error: line 145: Unknown tag: %python_provide: ERROR: python3-solv not recognized.`

pcreech:
This PR will be superseded by two other PRs breaking the individual items apart.

python-solv - https://github.com/theforeman/pulpcore-packaging/pull/938
libsolv - {}

Odilhao:
Can we close this @pcreech ?

Odilhao:
Closing this PR for now, we will soon rebuild everything with python 3.12.

PR #891, opened 2024-03-06
Update python39-aiohttp obsoletes to fix upgrades

pcreech:


pcreech:
I updated it to 4.0.0 to hopefully prevent this situation happening again in the future for older versions upgrading.

zjhuntin (approved):

PR #935, opened 2024-04-08
Exclude test directories in Pulp plugins

ehelms:


evgeni:
Would you mind rebasing this please?

Otherwise LGTM

PR #939, opened 2024-04-15
Provides for file and certtguard functionality

pcreech:
Since this functionality isn't going away, just being merged into pulpcore, pulpcore should now 'Provide' this functionality, to ensure proper upgrade paths.

### DIFF
--- a/packages/python-pulpcore/python-pulpcore.spec
+++ b/packages/python-pulpcore/python-pulpcore.spec
@@ -10,7 +10,7 @@
 
 Name:           python-%{pypi_name}
 Version:        3.49.3
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Pulp Django Application and Related Modules
 
 License:        GPLv2+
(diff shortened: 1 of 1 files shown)


packages/python-pulpcore/python-pulpcore.spec
```diff
@@ -126,9 +126,15 @@ Obsoletes:      python3-%{pypi_name} < %{version}-%{release}
 %if 0%{?rhel} == 8
 Obsoletes:      python39-%{pypi_name} < %{version}-%{release}
 %endif
+
 # Pulp-file and Pulp-certguard are now part of pulpcore
 Obsoletes:      python3.11-pulp-file < 1.16.0-1
+Provides:       python%{python3_pkgversion}-pulp-file
+Provides:       pulpcore-plugin(file)
```
evgeni: when it comes to the installer, we only "need" the `pulpcore-plugin(<name>)` notation:
https://github.com/theforeman/puppet-pulpcore/blob/21aa39e28f19b40bf328ebba514465fb9cfb9fe2/manifests/plugin.pp#L15

and I'd love those to be versioned, so
```suggestion
Provides:       python%{python3_pkgversion}-pulp-file = %{version}
Provides:       pulpcore-plugin(file) = %{version}
```

pcreech: based on "pulp-????-client" version for both of these plugins in pypi, agreed.

evgeni (approved):

PR #946, opened 2024-04-18
Add rhsm dependency for certguard

evgeni:


evgeni:
This used to be in `python-pulp-certguard`, but that's gone now.
https://github.com/theforeman/pulpcore-packaging/blob/5acfbf5b2d4a6b8f82aec41e71f437a660dd4a83/packages/python-pulp-certguard/python-pulp-certguard.spec#L35-L36

ekohl (approved):

PR #947, opened 2024-04-18
Update python-pulp-rpm to 3.25.3

sjha4:
Update pulp-rpm to 3.25.3 to pull in a needed fix: https://github.com/pulp/pulp_rpm/commit/e33b0e17d87fb990f3b97afcb7dc851e86aefe8b


Odilhao:
Can you also update here? @sjha4 https://github.com/theforeman/pulpcore-packaging/blob/0f439fab9fe397d03b0b0f57ec87dcd027329fea/automation/requirements.txt#L4

sjha4:
@Odilhao Updated..What does that do?

Odilhao:
> @Odilhao Updated..What does that do?

It's base for our automation, we keep it in sync with the RPM state, the plan is to have dependabot to manage that is the future

PR #960, opened 2024-04-22
Add python-rpm to requirement-check workflow

Odilhao:


Odilhao:
Will fix https://github.com/theforeman/pulpcore-packaging/actions/runs/8790110162/job/24121407582?pr=956

PR #969, opened 2024-04-24
Enable Legacy Hashes for EL9 builds of createrepo_c

Odilhao:


ianballou:
Solves https://projects.theforeman.org/issues/37332

ianballou (approved):
ACK from a Katello standpoint -- EL9 Katello users should be able to continue syncing repositories with SHA1 content.

PR #978, opened 2024-05-02
Declare pulp-cli-deb as dependency

Odilhao:


Odilhao:
This should be auto updated after https://github.com/pulp/pulp-cli-deb/issues/114

PR #984, opened 2024-05-08
Add workflow to bump package after dependabot merge

Odilhao:


.github/workflows/update-pulp-packages.yml
```diff
@@ -0,0 +1,79 @@
...
+        sudo curl --create-dirs -o /usr/local/bin/rpmdev-vercmp https://pagure.io/rpmdevtools/raw/79740e6f1881e399b0b4340a8090dd5adc91a4ea/f/rpmdev-vercmp
+        printf '#!/bin/bash\necho "$GIT_AUTHOR_NAME <$GIT_AUTHOR_EMAIL>"' | sudo tee /usr/local/bin/rpmdev-packager
+        sudo chmod +x /usr/local/bin/*
+    - name: Sort Updated Libs
+      if: github.event_name == 'push' && github.head_ref == 'rpm/develop'
+      run: git diff --unified=0 ${{github.event.before}} ${{github.sha}} automation/requirements.txt | grep -Po '(?<=^\+)(?!\+\+).*' > $PACKAGE_TO_UPDATE
+    - name: Check if package can be updated
+      if: github.event_name == 'push' && github.head_ref == 'rpm/develop'
+      run: |
+        set +e
```
ekohl: I'd strongly advise against big shell scripts in YAML. Create a separate shell script and run that. If it's limited to just GHA, store it in `.github`. That way you can modify it with proper syntax highlighting and tools like shellcheck.

.github/workflows/update-pulp-packages.yml
```diff
@@ -0,0 +1,79 @@
...
+          fi
+        done < $PACKAGE_TO_UPDATE
+      output:
+    - name: Open a PR
+      if: github.event_name == 'push' && github.head_ref == 'rpm/develop'
+      uses: peter-evans/create-pull-request@v6
+      with:
+        commit-message: "Update ${{ pkg }}" to ${{ NEW_VERSION }}""
+        branch: "bump_rpm/${{ pkg }}"
+        title: "Update ${{ pkg }}to ${{ NEW_VERSION }}"
```
ekohl:
```suggestion
        title: "Update ${{ pkg }} to ${{ NEW_VERSION }}"
```

.github/workflows/update-pulp-packages.yml
```diff
@@ -0,0 +1,79 @@
...
+              echo "Packaged $pkg RPM is newer than version in requirements"
+              exit 0
+          fi
+        done < $PACKAGE_TO_UPDATE
+      output:
+    - name: Open a PR
+      if: github.event_name == 'push' && github.head_ref == 'rpm/develop'
+      uses: peter-evans/create-pull-request@v6
+      with:
+        commit-message: "Update ${{ pkg }}" to ${{ NEW_VERSION }}""
```
ekohl: I don't think this is limited to a single pkg all the time. It can be 0, 1 or more.

ekohl (commented):
I'd still argue for an approach where `requirements.txt` isn't strict, but rather pins to ranges, for example using `pulp ~= 3.39.0` (which also allows installing 3.39.1). Then you'd install that and see if you need to update packaging. Otherwise `requirements.txt` goes out of sync with the rest of the repository all the time.

An additional benefit is that it can also work for stable branches: just pin to the .z stream.

In foreman-packaging we have:

* https://github.com/theforeman/foreman-packaging/blob/master/.github/workflows/bump_packages.yml

This builds up a list of updates to start. Each branch has a script to list the available updates. In practice that's https://github.com/theforeman/foreman-packaging/blob/rpm/develop/list_updatable_packages and https://github.com/theforeman/foreman-packaging/blob/deb/develop/scripts/list_updatable_packages.

For pulpcore-packaging, that could take https://github.com/theforeman/pulpcore-packaging/blob/rpm/develop/automation/_generate_deps.sh and format it in compatible way.

Then we have generated a matrix with jobs to update. You can see how we do that for RPMs here:

https://github.com/theforeman/foreman-packaging/blob/afb6040d279916cb79065468e719ea6495c163fc/.github/workflows/bump_packages.yml#L33-L36

For pulpcore-packaging you may want to take a different approach and group them, but in my experience with foreman-packaging it's better to have them split so you can review things individually.

Odilhao:
> I'd still argue for an approach where `requirements.txt` isn't strict, but rather pins to ranges, for example using `pulp ~= 3.39.0` (which also allows installing 3.39.1). Then you'd install that and see if you need to update packaging. Otherwise `requirements.txt` goes out of sync with the rest of the repository all the time.

We have [dependabot](https://github.com/theforeman/pulpcore-packaging/blob/rpm/develop/.github/dependabot.yml#L15) enabled now, and it's working to keep with z-streams of pulpcore and pulp-rpm, we need this due to VCRs on Katello that can get broken when we do major upgrades. 

I'll refactor later this week and move from being one big script inside of YAML to one standalone script.

I do have plans to revive the `_generate_deps.sh`, unfortunately right now we have packages that still use setup.py and other that don't use it anymore or never used, and the spec generation is not straightforward in the dependency side. For now we will only focus on getting the core components of pulp up to date, to later implement as Foreman is doing right now.

Odilhao:
@ekohl I think we are now in a better shape for another review. :smile:

Odilhao:
Let me tag @evgeni as well, I heard that he loves GH-actions

zjhuntin (approved):
Logic of this looks good to me!

Odilhao:
Merging and YOLO

PR #1029, opened 2024-06-18
Update python-pulp-cli and glue to 0.25.6

Odilhao:


packages/python-pulp-cli/python-pulp-cli.spec
```diff
@@ -6,7 +6,7 @@
 %global pypi_name pulp-cli
 
 Name:           python-%{pypi_name}
-Version:        0.25.4
+Version:        0.25.6
```
evgeni, line 9: The requests upper bound was changed back to <2.32 because there are issues with 2.32, you might want to change that

Odilhao: Oh nice. let me fix that

evgeni (approved):

PR #1074, opened 2024-07-31
Fix logic on json load to build matrix

Odilhao:
```
Invalid workflow file: .github/workflows/update-pulp-packages.yml#L32
The workflow is not valid. .github/workflows/update-pulp-packages.yml (Line: 32, Col: 9): Unrecognized named-value: 'matrix'. Located at position 1 within expression: matrix.package_name != '[]' .github/workflows/update-pulp-packages.yml (Line: 37, Col: 18): Unrecognized named-value: 'env'. Located at position 10 within expression: fromJson(env.matrix)
```

This PR will fix the logic error while loading the Json from `build_matrix.py`

### DIFF
--- a/.github/workflows/update-pulp-packages.yml
+++ b/.github/workflows/update-pulp-packages.yml
@@ -16,6 +16,8 @@ jobs:
     name: "Generate Package List"
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.head_ref == 'rpm/develop'
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
     - name: Sort Updated Libs
       if: github.event_name == 'push' && github.head_ref == 'rpm/develop'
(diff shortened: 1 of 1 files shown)


zjhuntin (approved):

.github/workflows/update-pulp-packages.yml
```diff
@@ -16,15 +16,17 @@ jobs:
...
     - name: Sort Updated Libs
       if: github.event_name == 'push' && github.head_ref == 'rpm/develop'
       run: git diff --unified=0 ${{github.event.before}} ${{github.sha}} automation/requirements.txt | grep -Po '(?<=^\+)(?!\+\+).*' > $PACKAGE_TO_UPDATE
     - name: Parse Package List
       run: ./build_matrix.py
     - name: Load Packages Matrix
-      run: |
-        echo "matrix=$(cat packages.json)" >> $GITHUB_ENV
+      id: set-matrix
+      run: echo "matrix=$(jq -c . < ./packages.json)" >> $GITHUB_OUTPUT
```
ekohl, line 29: Why is it needed to use `jq`? Wouldn't it be better to modify `build_matrix.py` to generate compact output?

PR #1076, opened 2024-07-31
Use needs.generate_package_list.outputs.matrix instead of steps.set-matrix.outputs.matrix

Odilhao:


zjhuntin (approved):
Looks good based on 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#needs-context
and https://github.com/theforeman/foreman-packaging/blob/master/.github/workflows/bump_packages.yml#L31

PR #1091, opened 2024-07-31
Rewrite build_matrix.py to set an output

ekohl:
Rather than using shell to set an output, this moves it into the matrix script itself. This is also done in foreman-packaging and works reliably.

It also reverts back to using include with matrix. This is because the output generated doesn't match a normal matrix definition. Normal matrices have multiple vectors and it runs for all possible combinations. You can also exclude some combinations, but more important here: you can include combinations. That's what build_matrix.py always generated: a list of combinations.

### DIFF
--- a/.github/workflows/update-pulp-packages.yml
+++ b/.github/workflows/update-pulp-packages.yml
@@ -27,10 +27,8 @@ jobs:
     - name: Sort Updated Libs
       run: git diff --unified=0 ${{github.event.before}} ${{github.sha}} automation/requirements.txt | grep -Po '(?<=^\+)(?!\+\+).*' > $PACKAGE_TO_UPDATE
     - name: Parse Package List
-      run: ./build_matrix.py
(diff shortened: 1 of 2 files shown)


.github/workflows/update-pulp-packages.yml
```diff
@@ -39,7 +37,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix: ${{ fromJson(needs.generate_package_list.outputs.matrix) }}
+      matrix:
+        include: ${{ fromJson(needs.generate_package_list.outputs.matrix) }}
```
Odilhao: This line got added with https://github.com/theforeman/pulpcore-packaging/commit/ee9da04cd5392907985023c8e093b0df9ca0d3f7 we are good to drop here

Odilhao (approved):

PR #1102, opened 2024-08-02
Add one check job for updates on automation/requirements.txt

Odilhao:


.github/workflows/requirement-check.yml
```diff
@@ -59,3 +59,21 @@ jobs:
...
+  check_resolution:
+    name: Check Package Installation
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/centos/centos:stream9
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
ekohl: I think 0 means the whole history, but here you only need a checkout which I think is the default:
```suggestion
```

Odilhao: Yes, I'm going to remove the checkout, we just need to check the current HEAD

ekohl (approved):
Would it make sense to make `rpm_list` depend on `check_resolution` (via `needs: check_resolution`)?

Odilhao:
> Would it make sense to make `rpm_list` depend on `check_resolution` (via `needs: check_resolution`)?

Yes, it does make sense to wait on `check_resolution`.

ekohl (approved):

ekohl (commented):
Last thought: would this workflow work if you also ran it when `.github/workflows/requirement-check.yml` itself changes? Sort of implementing CI for your CI.

Odilhao:
> Last thought: would this workflow work if you also ran it when `.github/workflows/requirement-check.yml` itself changes? Sort of implementing CI for your CI.

For that to happen we need to change this conditional

https://github.com/theforeman/pulpcore-packaging/blob/625443230d3d90e6fa810d12f1dbb9a33e05936a/.github/workflows/requirement-check.yml#L6

Let me send the change

ekohl:
> For that to happen we need to change this conditional

Exactly what I had in mind. And we immediately see it fail, so it's "working" :)

Odilhao:
```
 ERROR: Cannot install -r automation/requirements.txt (line 1) and pulp-glue==0.27.1 because these package versions have conflicting dependencies.
The conflict is caused by:
    The user requested pulp-glue==0.27.1
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
    pulpcore 3.49.15 depends on pulp-glue<0.26 and >=0.18.0
To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

I consider this a win, it's working!

PR #1130, opened 2024-08-12
Enable Auto Package Update

Odilhao:
This is the first stage of auto package updates, with this commit we will introduce the update for all packages required by [Pulpcore](https://github.com/pulp/pulpcore) and the plugins used by [Katello](https://github.com/Katello/katello).

This commit will not update the buildroot packages that are required to build all the packages, this will be done in a 2nd stage, using [pybuild-deps](https://github.com/bruno-fs/pybuild-deps), I'm in touch with @bruno-fs to fix some edge cases that I found when I tried to bundle libraries and buildroot dependencies at the same time.

### DIFF
--- a/.github/workflows/generate-package-list.yml
+++ b/.github/workflows/generate-package-list.yml
@@ -1,4 +1,4 @@
-name: Generate full list of packages
+name: Update Lib Packages
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


Odilhao:
@ekohl @evgeni :eyes:

ehelms (approved):
I am not a github actions expert, so I would appreciate a once-over from @evgeni or @ekohl. But I am also fine for you to merge and iterate as this is a very nice to have.

Odilhao:
I really wanna see what's going to break after we merge this.

Odilhao:
See, it's broken 

```
You have an error in your yaml syntax on line 42
```

PR #1144, opened 2024-08-22
Add task indices to pulpcore 3.49

sjha4:
Apllying the changes from https://github.com/pulp/pulpcore/pull/5368 to 3.49.

Patch being added is this: https://github.com/pulp/pulpcore/pull/5368.patch

Adding indices to the task table significantly improves task performance, however the change is in 3.54. We want this patch to be available on 3.49. 

Also need to ensure that patching between 3.49 and 3.54 doesn't break the eventual upgrade from 3.49 to 3.54. The migration task is not idempotent and ensuring the same name of migration file in theory should prevent django from running it twice.

More details: https://issues.redhat.com/browse/SAT-27211


sjha4:
@evgeni @Odilhao Need your beautiful eyes on this one.. 👀

ianballou (approved):
Working great, I installed the [RPM](https://download.copr.fedorainfracloud.org/results/@theforeman/pulpcore-nightly-staging-scratch-da328d1c-57ea-51d5-9d36-a6e03c34e0e1/rhel-9-x86_64/07935262-python-pulpcore/python3.11-pulpcore-3.49.17-2.el9.noarch.rpm) and then ran the migrations. The migration in the patch ran cleanly. I then upgraded (via `pip3.11`) to the following Pulp versions:
```
      "versions": {
        "deb": "3.3.1",
        "rpm": "3.27.1",
        "core": "3.58.0",
        "file": "3.58.0",
        "ostree": "2.4.3",
        "python": "3.12.2",
        "ansible": "0.22.1",
        "certguard": "3.58.0",
        "container": "2.21.0"
      }
```
and also ran the migrations. After applying the migrations, fixing the systemd service files, and restarting everything, I was still running a working environment.

sjha4:
@evgeni @pcreech This one needs an ack with merge superpowers..

pcreech:
Just pointing out:

With Django, this approach is dangerous and has potential knock on effects for upgrades if we don't get this right.  We are going to want to make sure we test this and pay close attention to it.  It's an approach that once we open this box can't be undone and could increase the support costs in the long run.

I've chatted with the pulpcore team, and in this case, this particular migration should have minimal risks, comparatively, for this situation.  but we should not commit to this in the long run unless we've vetted it.

If possible, i'd like to see if we can get a lot of testing on the attached rpms before we merge this, to ensure the issues are resolved as expected and we identify any upgrade pathways that could have an issue.

pcreech:
After conversations with all parties involved, I'm satisfied we've had the due diligence done. ACK, merging.

PR #1151, opened 2024-09-02
Pin pulp-glue-deb and pulp-cli-deb updates to be on the same PR

Odilhao:


.github/dependabot.yml
```diff
@@ -19,6 +19,11 @@ updates:
           patterns:
             - "pulp-cli"
             - "pulp-glue"
+       pulp-deb-deps:
```
quba42: This is just naming, but I think it is technically more accurate/more consistent with other naming elsewhere:
```suggestion
       pulp-cli-deb-deps:
```

quba42 (approved):
I have one cosmetic suggestion. Take it or leave it, everything LGTM, and I think this makes sense.

.github/dependabot.yml
Odilhao: I'll take that

quba42 (approved):

PR #1243, opened 2024-09-12
Drop charset-normalizer requirement from python-aiohttp

Odilhao:
this was removed with 3.9.z release

https://github.com/aio-libs/aiohttp/blob/v3.9.5/setup.cfg#L51


evgeni (approved):
https://github.com/aio-libs/aiohttp/commit/79c97f70111418c34473852ce0938172f0f43802

PR #1288, opened 2024-09-20
Update requirements to Pulpcore 3.61.0

Odilhao:


ianballou (commented):
Is nightly going to start running with these versions? Asking since we can't upgrade our client bindings quite yet, we are waiting for https://github.com/Katello/katello/pull/11058 to merge first so the VCRs can get recorded together.

It might not be an issue to start running with these Pulp versions in nightly, but there's always a chance that the new Pulp won't work with the old client bindings.

Odilhao:
Not yet, after we merge this PR, we will get one run of https://github.com/theforeman/pulpcore-packaging/blob/rpm/develop/.github/workflows/update-pulp-packages.yml and it will create PRs for each lib that we are updating, we will only break nightly after we merge the package PRs, we would benefit with merging this today to ensure that we will have a run of https://github.com/theforeman/pulpcore-packaging/blob/rpm/develop/.github/workflows/generate-package-list.yml with the new pulp version, we might be able to get some new lib update PR.

ianballou (approved):
Works for me then!

Issue #1244, opened 2024-09-12
Add package for google-cloud-storage


dmaes:
We are running pulp on GCP, and would like to use Google Storage Buckets as storage backend. Would it be possible to package google-cloud-storage? It seems to be required by `django-storages` to be able to use Google Storage.

Thanks in advance.

ishan4421:
Hi @dmaes , You can surely do it. I used GCP for my NLP project but it wad not related to Django. I am attaching a chatgpt chat for it. I think it will work. I hope you have GCP json file. I hope this will work. 

 https://chatgpt.com/share/66e4859e-e25c-8006-818b-211c7af0fc8d

Odilhao:
Hey @dmaes , looking at the [required](https://github.com/googleapis/python-storage/blob/main/setup.py#L30) packages atm we only have requests packaged, we would need to package 5 new packages. I'll try to take a look next week, we can draft the PRs, do you know if that's the only requirement for pulp to work with GCP Buckets?

dmaes:
Hi @ishan4421, thanks for the "how", but that was not necessarily what I was asking, I figured that out already. My question was if it would be possible to make and add a `python-google-cloud-storage` RPM for the `google-cloud-storage` python package, since that is currently not present in this repository, but is required to actually be able to use Google cloud storage.

dmaes:
HI @Odilhao, missed your comment while writing mine. As far as I understand the Django storages docs and the dependencies for [`django-storages[google]`](https://github.com/jschneier/django-storages/blob/b85d0214dada173059268d85fe5d318215480c08/pyproject.toml#L57) correctly, `google-cloud-storage`, allong with it's dependencies, should be the only requirement. But I will double-check on monday, and test it out with a pulp venv/pip install.

dmaes:
Just tested, and can confirm that `google-cloud-storage`, along with it's dependencies, is enough.

dmaes:
@Odilhao I tried creating a PR, not sure if it's any good, especially with the git annex stuff.

dmaes:
@Odilhao thanks for the help. Everything running smoothly on GCS buckets now.

PR #1332, opened 2024-09-23
add packages for google-cloud-storage

dmaes:
An attempt at a PR for #1244 . Not entirely sure if I did it correctly, especially with the git-annex stuff, please let me know if I need to fix something.


Odilhao:
Thanks for adding the packages @dmaes , looking at the files I can see that the script generated the `scl` bits that we don't use due to EL7 being deprecated, we need to remove all scl reference and google packages always explode the build requires with the protobuf section, I'll do some review later today and trim the spec files. 

We also need some e-mail in the specfile changelog, atm it's only showing the date.

dmaes:
@Odilhao thanks for taking a look. I fixed the changelogs already.

Odilhao:
ok to test

Odilhao:
@dmaes I'm finishing locally the addition of 2 packages `rsa` and `cachetools` I believe that after adding those 2 we will have all the necessary packages.

Odilhao (approved):

Odilhao:
Repoclosure failed because we don't have the packages published, all builds passed, merging.

PR #1356, opened 2024-10-07
Drop EL8 from Pulpcore Nightly

Odilhao:


package_manifest.yaml
```diff
@@ -16,20 +16,11 @@ packages:
     copr_projects:
       - "{{ hostvars['pulpcore-copr'] }}"
     koji_tags:
```
ekohl: I don't think koji_tags is used at all anymore. Did we forget to clean this up previously?

package_manifest.yaml
```diff
@@ -16,20 +16,11 @@ packages:
...
-      - name: pulpcore-{{ pulpcore_version }}-el8
-        dist: '.el8'
       - name: pulpcore-{{ pulpcore_version }}-el9
         dist: '.el9'
     repoclosure_target_repos:
-      el8:
-        - "el8-pulpcore-{{ pulpcore_version }}-staging"
       el9:
         - "el9-pulpcore-{{ pulpcore_version }}-staging"
-    repoclosure_lookaside_repos:
```
ekohl, line 28: You should have kept this line in

package_manifest.yaml
```diff
@@ -47,7 +38,6 @@ copr_projects:
     core_modules:
       - 'ruby:2.7'
```
ekohl: This looks unused now

ekohl (changes requested):

package_manifest.yaml
Odilhao: Yes, this one was left behind

package_manifest.yaml
```diff
@@ -15,21 +15,10 @@ packages:
     build_package_use_koji_build: true
```
ekohl, line 15: Is this also still relevant? I think this was made redundant when we switched to copr.

ekohl (approved):
:+1: once we've dropped EL8 from Foreman

PR #1359, opened 2024-10-09
add psycopg_c

evgeni:


evgeni:
Probably should add a `Requires` in `pulpcore`:
https://github.com/theforeman/pulpcore-packaging/blob/3892e020273adef1ad073b2ff547209dba0c33d7/packages/python-pulpcore/python-pulpcore.spec#L101-L102

comps/comps-pulpcore-el8.xml
```diff
@@ -204,6 +204,7 @@
       <packagereq type="default">python3.11-protobuf</packagereq>
       <packagereq type="default">python3.11-ptyprocess</packagereq>
       <packagereq type="default">python3.11-psycopg</packagereq>
+      <packagereq type="default">python3.11-psycopg_c</packagereq>
```
Odilhao, line 207: No need to add EL8 on nightly , if we want to backport this to 3.49 we can add there, we will kill EL8 this week or on the next.

Odilhao (approved):

comps/comps-pulpcore-el8.xml
evgeni: Well, yeah. But today we have el8, so we can also build it and kill all next week :)

Odilhao:
We need this now https://github.com/theforeman/pulpcore-packaging/pull/1361

PR #1363, opened 2024-10-09
Add psycopg_c as requirement for Pulpcore 3.49

Odilhao:


Odilhao:
another PR for you @evgeni

packages/python-pulpcore/python-pulpcore.spec
```diff
@@ -102,7 +102,9 @@ Requires:       python%{python3_pkgversion}-opentelemetry_instrumentation_wsgi <
 Requires:       python%{python3_pkgversion}-protobuf >= 4.21.1
 Requires:       python%{python3_pkgversion}-protobuf < 4.25.4
 Requires:       python%{python3_pkgversion}-psycopg >= 3.1.8
-Requires:       python%{python3_pkgversion}-psycopg <= 3.1.18
```
evgeni, line 105: You're changing the upper bound -- is that OK?

Odilhao: Looks like we have a problem https://github.com/pulp/pulpcore/blob/3.49.20/requirements.txt#L34 let me ask Pulp and see if they can bump this.

evgeni: https://github.com/pulp/pulpcore/blob/3.49/requirements.txt#L33 says no

Odilhao: Yep, asked to see if we can get a backport to 3.49 for this

Odilhao: Pulpcore 3.49.21 was just release with this change https://github.com/pulp/pulpcore/blob/3.49.21/requirements.txt#L33

PR #1412, opened 2024-10-30
Filter Packages with .post in the version

Odilhao:


ekohl (commented):
I don't think this is right because you'll get updates from x.y.z to x.y.z, but you'll miss updates. Should we just accept that some upstreams do weird things?

Odilhao:
Yes, I was looking on how post work in some packages and it's just weird how some packages use only for ci related and other release minor changes.

ekohl:
Yes, but I think I saw a PR bump from x.y.z to x.y.z.post1 and I'm unsure what happens then

Odilhao:
It's probably the bracex PR https://github.com/theforeman/pulpcore-packaging/pull/1338 it's just CI stuff.

https://github.com/facelessuser/bracex/compare/2.5...2.5.post1

PR #1492, opened 2024-12-16
Add patch to Mark md5 usage as usedforsecurity=False on Pulpcore

Odilhao:


Odilhao:
@ianballou @ehelms any concern on adding this patch to nightly builds?

ehelms:
Why does it need to be included here? I think that would help me give you an answer.

Odilhao:
This patch was never part of pulpcore upstream codebase, it's a patch that we need to add to the specfile.

packages/python-pulpcore/python-pulpcore.spec
```diff
@@ -191,6 +191,9 @@ done
 
 
 %changelog
+* Mon Dec 16 2024 Odilon Sousa <osousa@redhat.com> - 3.63.6-2
+- Add patch to Mark md5 usage as usedforsecurity=False on Pulpcore
```
evgeni: You need to actually reference that patch :)

Odilhao: Ok, done.

evgeni: no, I meant you need `Patch0` in the spec and a `%patch0` macro call

Odilhao: Wait what, I need a break, added the source, since we already have `%autosetup -p1 -n %{pypi_name}-%{version}` we are good in the macro section.

Odilhao: Looks like we got the patch applied

```
+ cd pulpcore-3.63.6
+ /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
+ /usr/bin/cat /builddir/build/SOURCES/0001-Mark-md5-usage-as-usedforsecurity-False.patch
+ /usr/bin/patch -p1 -s --fuzz=0 --no-backup-if-mismatch
+ rm -rf pulpcore.egg-info
+ sed -i 's/psycopg\[binary\]/psycopg/' requirements.txt
+ RPM_EC=0
```

evgeni: ah yes, autosetup does that.

ehelms (approved):

PR #1875, opened 2025-04-22
pulpcore-selinux: Use Python 3.12 as used by Pulpcore 3.73+

### LINKED ISSUE #81: /run/pulpcore-api.sock is blocked for httpd
After upgrading Katello to Pulpcore 3.73, this denial is popping up:

```
type=AVC msg=audit(1745189293.472:4348): avc:  denied  { connectto } for  pid=39529 comm="httpd" path="/run/pulpcore-api.sock" scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=unix_stream_socket permissive=0
```

Jira: https://issues.redhat.com/browse/SAT-33154

evgeni:


packages/pulpcore-selinux/pulpcore-selinux.spec
```diff
@@ -16,7 +16,7 @@ BuildRequires:  checkpolicy
 BuildRequires:  selinux-policy-devel
 BuildRequires:  systemd
 Requires:       selinux-policy >= %{_selinux_policy_version}
-Requires:       pulpcore
+Requires:       pulpcore >= 3.73
```
ekohl: Should we require it via `python%{pulpcore_python}dist(pulpcore)` so in the future the Python version will always match? We also have the `fixfiles` invocations and I wonder if we could match those. Perhaps via a dedicated macro that we use both here and the `fixfiles` calls.

evgeni: `fixfiles` only knows about (real) rpm names (it does `rpm -q` under the hood), so it has to be `pythonX.Y-pulpcore`.
given we have the python version as a macro, also using `pythonX.Y-pulpcore` here would also work (and would have exposed the issue earlier!)

ekohl (changes requested):
There's also a `fixfiles` in %post` (which I can't reply to inline), but :+1: for the rest.

ekohl (approved):

PR #2065, opened 2025-07-06
Refactor find package.py

Odilhao:
## Summary

Refactored the `parse_package_list` function in `find_package.py` to improve maintainability and added comprehensive test coverage.

### Changes Made

**Refactoring:**
- Replaced 40+ elif statements with organized data structures (prefix_removals, package_mappings, lowercase_packages, pattern transformations)
- Improved code readability and made adding new package transformations much easier
- Fixed pattern matching logic to be more selective and accurate
- All existing functionality preserved with no breaking changes

**Testing:**
- Added comprehensive test suite with 9 tests covering all transformation scenarios
- Included integration test with real-world data (191 packages)
- Added test documentation and easy-to-use test runner
- All tests pass successfully

### Benefits

- **Maintainable**: Adding new package transformations is now straightforward
- **Reliable**: Comprehensive test coverage ensures correct behavior
- **Documented**: Clear test documentation explains all transformation rules
- **Verified**: Integration test validates compatibility with production data

The refactored code handles the same package name transformations as before but with significantly better organization and maintainability.


run_tests.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
JacobCallahan: This file shouldn't be needed anymore now that you're using pytest

test/test_integration.py
```diff
@@ -0,0 +1,268 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+
+# Add the parent directory to the path so we can import find_package
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
JacobCallahan: not a blocker, but since this is present in both test files, this could be added to a conftest.py file in the form of a session-scoped auto-use fixture.

Odilhao: Moved the oneline to `conftest.py`

with

```
#!/usr/bin/env python3

import sys
import os
import pytest


@pytest.fixture(scope="session", autouse=True)
def setup_import_path():
    """Add the parent directory to sys.path so we can import find_package module."""
    parent_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
    if parent_dir not in sys.path:
        sys.path.insert(0, parent_dir) 
```

pcreech (approved):
Locally, the tests seem to result in success.  The base functionality of find_packages.py also appears to be retained.  LGTM from a general correctness standpoint.

find_package.py
```diff
@@ -6,68 +6,102 @@
...
+        'aiohttp_socks': 'aiohttp-socks',
+        'pyasn1_modules': 'pyasn1-modules',
+        'pydantic_core': 'pydantic-core',
+        'flit_core': 'flit-core',
+        'poetry_core': 'poetry-core',
+        'poetry_plugin_export': 'poetry-plugin-export',
+    }
+    
+    # Packages that need to be lowercased
+    lowercase_packages = {
```
decko, line 35: There's any package name that needs to keep its uppercase characters? If isn't, why not simply ask for lowercase all package names and remove this dict here?

find_package.py
```diff
@@ -6,68 +6,102 @@
...
+        'Pygments': 'pygments',
+        'PyJWT': 'pyjwt',
+        'RapidFuzz': 'rapidfuzz',
+        'SecretStorage': 'secretstorage',
+    }
+    
+    # Pattern-based transformations
+    def apply_pattern_transformations(name):
+        # OpenTelemetry packages: replace - with _
+        if name.startswith('opentelemetry'):
```
decko, line 55: That dict idea could also be extended a little to allow rules like this:
{"opentelemetry-": {"rule": "starts_with", "from": "-", "to": "_"}}.

This could avoid all those ifs here.

Odilhao: The main problem is that we have packages that need to keep the `-`, it's python packaging in the end of the day, everyone is doing it's own thing.

Odilhao: I'm going to address this in another PR.

PR #2319, opened 2026-01-07
Bump pulp-deb from 3.7.0 to 3.8.0 in /automation

quba42:


quba42:
Background: We need pulp_deb 3.8.0 so we can add the alternate content sources (ACS) feature for deb to Katello, here: https://github.com/Katello/katello/pull/11567

pulp_deb dependencies have not changed between 3.7.0 and 3.8.0, so this should not be a problem. We will of course also need the newer client.

Question: Am I correct in thinking that bumping the version in `automation/requirements.txt` means some automation will automatically create a PR to update the package itself?

Odilhao (approved):

Odilhao:
This is the correct place to update @quba42 , after you change from Draft we can merge.

ianballou:
We typically pair bumps like this with a bump in Katello. Oftentimes with Pulp plugin y-updates, VCRs need to be rerecorded in Katello and things like smart proxy syncing need to be tested. Were those aready tested with Pulp Deb 3.8?

PR #2766, opened 2026-07-20
Remove obsolete packages for EL10 transition

zjhuntin:
## Summary
- Deletes 63 package directories and 2 orphaned directories (libsolv0, subscription-manager)
- Removes 44 entries from `package_manifest.yaml`
- Removes 48 entries from `comps/comps-pulpcore-el9.xml`
- Drops stale `Requires` from 8 dependent specs:
  - python-aioredis: drop async-timeout (+ Release bump)
  - python-socks: drop async-timeout (+ Release bump)
  - python-tablib: drop xlrd, xlwt (+ Release bump)
  - python-pulpcore: drop importlib-metadata (+ Release bump)
  - python-pulp-rpm: drop importlib-resources (+ Release bump)
  - python-opentelemetry_api: drop importlib-metadata (+ Release bump)
  - python-build: drop importlib-metadata (+ Release bump)
  - python-flake8: drop importlib-metadata (+ Release bump)
- Removes python-aioredis entirely (dead python3.11 package, already obsoleted)

**Merge after** #2765 (the Obsoletes-only PR) — this PR may cause RPM pipeline failures while builds settle, which is expected.

## Context
- SAT-45013: EL10 transition cleanup
- Split from #2751 and #2764 per review feedback
- python-tomli removal deferred to a future PR (~36 dependent specs need updating)

## Test plan
- [ ] Verify `package_manifest.yaml` is valid YAML
- [ ] Verify `comps/comps-pulpcore-el9.xml` is valid XML
- [ ] Verify no remaining specs have dangling Requires on deleted packages
- [ ] RPM pipeline — failures expected and acceptable for this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2765, opened 2026-07-20
Add Obsoletes for EL10 package removals

zjhuntin:
## Summary
- Bumps `pulpcore-obsolete-packages` from 1.4.0 to 1.5.0
- Adds Obsoletes entries for 63 packages being removed during the EL10 transition
- 25 new python3.11 Obsoletes for packages that were still on python3.11
- 25 new python3.12 Obsoletes for packages that had transitioned to python3.12
- Bumps existing python3.11-aioredis Obsoletes version to cover Release -7

This PR is **safe to merge independently** — it only adds Obsoletes directives to force-remove old packages during upgrades. No package directories are deleted, no manifest/comps changes, no pipeline risk.

The actual package directory deletions, manifest/comps cleanup, and dependent spec Requires fixes will follow in a separate PR (#2766).

## Context
- SAT-45013: EL10 transition cleanup
- Split from #2751 and #2764 per review feedback — Obsoletes-only changes land first

## Test plan
- [ ] Verify spec syntax is valid (`rpmlint`)
- [ ] Verify Obsoletes version-release values are >= last shipped NVR for each package
- [ ] RPM pipeline builds successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2768, opened 2026-07-22
Restore python-importlib-metadata and python-importlib-resources

Odilhao:
Nightly staging repoclosure broke after #2765/#2766 (EL10 obsolete-package
cleanup) merged:

```
package: python3.12-bandersnatch-6.6.0-1.el9.noarch from el9-pulpcore-nightly-staging
  unresolved deps (1):
    python3.12dist(importlib-metadata)
package: python3.12-opentelemetry_api-1.40.0-2.el9.noarch from el9-pulpcore-nightly-staging
  unresolved deps (1):
    (python3.12dist(importlib-metadata) < 8.8 with python3.12dist(importlib-metadata) >= 6)
package: python3.12-pulp-rpm-3.35.3-2.el9.noarch from el9-pulpcore-nightly-staging
  unresolved deps (1):
    (python3.12dist(importlib-resources) >= 6.4 with python3.12dist(importlib-resources) < 6.5)
```

## Root cause

`python-importlib-metadata` and `python-importlib-resources` were removed on
the theory that they're stdlib-replaceable backports, no longer needed on
Python 3.11+/3.12. That's true for the packages we checked at review time
(pulpcore, python-build, python-flake8, python-tablib, python-socks — each
either dropped the dependency upstream or gates it behind a `python_version`
marker that's false on 3.12).

It is **not** true for three active consumers, which declare it as an
unconditional dependency in their own PyPI metadata (no marker at all):

- `opentelemetry-api` 1.40.0: `importlib-metadata<8.8.0,>=6.0`
- `bandersnatch` 6.6.0: `importlib_metadata` (unbounded)
- `pulp-rpm` 3.35.3: `importlib-resources~=6.4.0`

`bandersnatch` never had a hand-written `Requires` on this at all — its
dependency comes purely from RPM's auto-generated python dist-based
`Requires` (`python3.12dist(importlib-metadata)`), derived from its wheel
metadata. That's satisfied only by the actual PyPI backport package's
`Provides` — stdlib `importlib.metadata` does not supply it. Deleting the
package (not just the manual `Requires` lines in the other two specs)
removed the only provider.

## Fix

Restores both packages at their last-shipped NVR, each in its own commit:
- `python-importlib-metadata` 6.0.1-7 (was 6.0.1-6 before removal; bumped
  since it's a re-add)
- `python-importlib-resources` 6.4.5-3 (was 6.4.5-2 before removal; bumped
  since it's a re-add)

Also flagged on #2765 (still open, not yet merged) that its
`Obsoletes: python3.12-importlib-metadata < 6.0.1-6` and
`Obsoletes: python3.12-importlib-resources < 6.4.5-2` lines need to be
dropped entirely — otherwise the restored packages here would immediately
self-obsolete once that PR lands.

The other 44 packages removed in #2766, and the 5 other dependent-spec
Requires drops (build, flake8, tablib, socks, pulpcore), were re-verified
against upstream metadata and are correct — repoclosure only flagged these
two.


zjhuntin (approved):
LGTM — clean hotfix, NVR bumps are correct, comps and manifest placements are right, and the PR description is excellent on the root cause.

A couple of minor things worth a follow-up (not blocking):

- **`python-importlib-resources` and `python-zipp` `%files`**: Both are missing `%license` / `%doc` macros. `python-importlib-metadata` has them; the other two don't. With `%pyproject_install` the dist-info carries the license so it's not a build blocker, but rpmlint will flag it. Worth adding in a subsequent bump.

- **PR #2765 Obsoletes interaction**: The description notes `Obsoletes: python3.12-importlib-metadata < 6.0.1-6` and `Obsoletes: python3.12-importlib-resources < 6.4.5-2` in #2765. With restored NVRs at 6.0.1-7 and 6.4.5-3 those bounds wouldn't actually self-obsolete (7 > 6, 3 > 2), but worth confirming the intent is to drop those lines entirely rather than update the bounds, before #2765 lands.

PR #2769, opened 2026-07-22
Add EL10 buildroot infrastructure for pulpcore-packaging

zjhuntin:
## Summary

- Creates `comps/comps-pulpcore-el10.xml` based on the EL9 equivalent (identical package list — both use python3.12)
- Adds `rhel_10: '10'` variable and `rhel-10-x86_64` chroot to `pulpcore-copr` in `package_manifest.yaml` with the same python3.12 buildroot packages as EL9
- Adds EL10 `repoclosure_target_repos`/`repoclosure_lookaside_repos` entries to the `packages` vars section, using `el10-baseos`, `el10-appstream`, and `el10-crb` as lookaside repos
- Adds a `pulpcore-staging-repoclosure-el10` host to the `repoclosures` section

Relates: [SAT-47787](https://redhat.atlassian.net/browse/SAT-47787)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Odilhao (approved):
Reviewed.

## comps/comps-pulpcore-el10.xml

Diffed byte-for-byte against the current `comps/comps-pulpcore-el9.xml` — identical, confirming the PR's claim. Makes sense given both target python3.12.

## package_manifest.yaml

The `rhel_10`/chroot/repoclosure additions are a clean structural mirror of the existing `el9` blocks, with one deliberate difference I dug into before approving:

- EL9's `rhel-9-x86_64` chroot buildroot lists `python3.12-rpm-macros`; the new EL10 chroot lists `python-srpm-macros` instead — not a pure `9→10` substitution. Checked this isn't a copy-paste slip: on RHEL 10, python3.12 becomes the **default** system Python (unlike RHEL 9, where it's an additional non-default interpreter installed alongside 3.9/3.11). The versioned `python3.12-rpm-macros` subpackage only exists for the non-default-interpreter case; once 3.12 is the default, the generic `python-srpm-macros` package is the right one for SRPM-level macros. This is the correct adaptation for EL10's packaging model, not a bug.
- `el10-baseos`/`el10-appstream`/`el10-crb`/`el10-foreman-plugins-nightly` repo IDs aren't defined anywhere in this repo (same is true for the existing `el9-*` ones — they're not in `mock/` or anywhere else tracked here), so I can't verify their existence from inside the repo. They exactly mirror the naming convention of the already-working `el9-*` entries, so I'm treating that as strong evidence they're provisioned correctly server-side (Jenkins), but flagging since it's the one part of this PR I couldn't independently confirm.

CI green, mergeable. Approved.

PR #2835, opened 2026-07-28
Bump libcomps release for EL10 rebuild verification

Odilhao:
## Summary
- EL10 rebuild (theforeman/el10_rebuild#15) — bump Release for libcomps to produce a real, verifiable build for both EL9 and EL10.
- Split out from the tier4 wave 1 bundled PR into its own PR since it needed real spec changes, not just a release bump.
- Trimmed the spec to only target el9/el10, using CentOS Stream 9/10's own `libcomps.spec` (`gitlab.com/redhat/centos-stream/rpms/libcomps`) as a structural reference:
  - Dropped SCL support and the python36 (EL8-only) build variant — both dead code paths in this repo now.
  - Adopted `%cmake_build`/`%cmake_install`/`%ldconfig_scriptlets` instead of the old mixed `%make_build`/manual scriptlet fallback.
  - Kept the `python3.12-libcomps` package name (unlike CS's unversioned `python3-libcomps`) since `python-pulp-rpm` explicitly `Requires: python3.12-libcomps >= 0.1.23`.
  - Kept building the `python39` bindings on el9 (its default python, needed by `dnf`) — not present in CS10 since that's an el10-only spec.
  - Added `Obsoletes: python3-libcomps` for el10 (matching CS10's naming for the python bindings subpackage) — el10's default python is also 3.12 (unlike el9's 3.9), so the OS-provided `python3-libcomps` would otherwise own the same site-packages files as ours.
- **Real bug found and fixed in `%check`:** the previous spec ran a bare `make test` inside `build-py3`, but the `%cmake` macro actually configures into a nested `build-py3/redhat-linux-build` directory — so `make test` there always hit "No rule to make target 'test'". This was masked on el9 by an unrelated `%if 0%{?rhel} != 9` skip, and only surfaced when el10 hit the same bug with no skip guard. Fixed by using `%cmake_build --target test` / `--target pytest`, which resolves the build directory correctly — confirmed via `obal scratch` that `%check` now actually executes and passes on both el9 and el10 (previously it either silently skipped or failed).
- Verified locally with `obal scratch libcomps` before pushing — both chroots green, with tests actually running this time.

## Test plan
- [ ] Jenkins/COPR CI green on rhel-9-x86_64 (with %check actually executing)
- [ ] Jenkins/COPR CI green on rhel-10-x86_64 (with %check actually executing)

### DIFF
--- a/packages/libcomps/libcomps.spec
+++ b/packages/libcomps/libcomps.spec
@@ -1,42 +1,17 @@
 %global python3_pkgversion 3.12
 %global __python3 /usr/bin/python3.12
 
-# explicitly define, as we build on top of an scl, not inside with scl_package
-%if 0%{?scl:1}
-%global scl_prefix %{scl}-
-%global python3_sitearch /opt/theforeman/tfm-pulpcore/root/usr/lib64/python3.8/site-packages/
-%global python3_version %python38python3_version
-%global __os_install_post %python38_os_install_post
-%global __python_requires %python38_python_requires
-%global __python_provides %python38_python_provides
-%global __python3 %python38__python
-%endif
-
-# Always build Python3 bindings
-%bcond_without python3
-
-# Our EL8 buildroots default to Python 3.8, but we also need a 3.6 build of libcomps
-# to make dnf happy
-%if 0%{?rhel} == 8
-%bcond_without python36
-%else
-%bcond_with python36
-%endif
-
-# Our EL9 buildroots default to Python 3.9, we also need a 3.9 build of libcomps
-# to make dnf happy
+# RHEL9's default python is 3.9, and dnf needs its own libcomps bindings
+# against that interpreter, separate from our python3.12 build.
 %if 0%{?rhel} == 9
 %bcond_without python39
 %else
 %bcond_with python39
 %endif
 
-# Never build docs by default
-%bcond_with doc
-
 Name:           libcomps
 Version:        0.1.23
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Comps XML file manipulation library
 
 License:        GPLv2+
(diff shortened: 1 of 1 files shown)


packages/libcomps/libcomps.spec
```diff
@@ -1,42 +1,17 @@
...
 %bcond_with python39
 %endif
 
-# Never build docs by default
-%bcond_with doc
-
 Name:           libcomps
 Version:        0.1.23
-Release:        1%{?dist}
+Release:        3%{?dist}
```
zjhuntin: Only needs one bump, I think.

Odilhao: Good catch, fixed — Release is now 2, not 3.

zjhuntin (commented):
## Review Comments

Two items worth clarifying before merge:

### 1. `python3-libcomps` in el10 comps vs. the new Obsoletes

`comps-pulpcore-el10.xml:28` lists `python3-libcomps` as a `<packagereq>`, but the spec now adds `Obsoletes: python3-libcomps` on EL10. This means the comps group explicitly requests a package that the repo's own `python3.12-libcomps` obsoletes. Could this cause confusing dep-solver behavior or warnings during repo composition / repoclosure? Should this entry be dropped from the el10 comps, or is it intentionally kept so the Obsoletes pulls `python3.12-libcomps` in via the obsoletion chain?

Same question applies to `python3-createrepo_c` — on EL10 where python3.12 is the system python, is there a similar file-conflict situation as `python3-libcomps` that would need an Obsoletes? (Or is that handled elsewhere?)

### 2. Duplicate entries in el10 comps

Since `comps-pulpcore-el10.xml` is a new file, it's a good opportunity to clean up the duplicates inherited from el9:

- `python3.12-createrepo_c` appears twice (lines 27 + 95)
- `python3.12-libcomps` appears twice (lines 178 + 179)
- `python3.12-solv` appears twice (lines 299 + 300)

These are pre-existing in el9, so not introduced by this PR — just a nice-to-have cleanup since the file is being created from scratch.

zjhuntin (approved):

PR #2950, opened 2026-08-24
Add Obsoletes/Provides for python3-cryptography and python3-requests on RHEL 10

### LINKED ISSUE #2949: Add Obsoletes/Provides for python3-cryptography on python-cryptography (EL10 conflict)
## Problem

Installing Satellite container images on RHEL 10 fails during `satellitectl pull-images`'s `pre_install` role with an RPM transaction conflict:

```
Unknown Error occurred: Transaction test error:
  file /usr/lib64/python3.12/site-packages/cryptography/... from install of
  python3-cryptography-43.0.0-4.el10.x86_64 conflicts with file from package
  python3.12-cryptography-46.0.7-2.el10pc.x86_64
```

(full file list truncated — dozens of overlapping paths under `cryptography/`)

## Root cause

[`foremanctl`'s `pre_install` role](https://github.com/theforeman/foremanctl/blob/master/src/roles/pre_install/tasks/main.yaml#L27) requests the package **by the plain, unversioned name** `python3-cryptography`:

```yaml
- name: Install other dependencies
  ansible.builtin.package:
    name:
      - bash-completion
      - python3-cryptography
      - python3-libsemanage
      - python3-psycopg2
      - python3-requests
      - python3-requests-oauthlib
```

On RHEL 10, `python3` == `python3.12`, and `python-cryptography.spec` here in pulpcore-packaging builds under the versioned name `python312-cryptography` (via `%{python3_pkgversion}`). Both names resolve to the same physical path (`%{python3_sitearch}` == `/usr/lib64/python3.12/site-packages/cryptography/`), so when a host already has `python312-cryptography` installed (pulled in normally via another dependency chain) and then also picks up a stale `python3-cryptography` build from elsewhere (e.g. the Satellite compose), dnf reports a hard file conflict rather than resolving it — RPM doesn't compare versions across differently-named packages.

There's already a precedent for exactly this in this repo: [`python-ldap.spec`](https://github.com/theforeman/pulpcore-packaging/blob/rpm/develop/packages/python-ldap/python-ldap.spec) carries inline

```spec
Obsoletes: python3-pyldap < 3
Provides:  python3-pyldap = %{version}-%{release}
Provides:  python3-pyldap%{?_isa} = %{version}-%{release}
```

on the package itself (not in the centralized `pulpcore-obsolete-packages`, which is reserved for retired `python3.9-*`/`python3.11-*`-style version transitions, not for the currently-active `python312-*` build claiming an unversioned BaseOS name).

## Ask

Add the same inline pattern to `python-cryptography.spec`, **gated to RHEL/CentOS Stream 10 only** — on RHEL 9, `python3` is 3.9 and `python3-cryptography` there is a distinct, legitimate package that shouldn't be touched:

```spec
%if 0%{?rhel} >= 10
Obsoletes: python3-cryptography < <appropriate version>
Provides:  python3-cryptography = %{version}-%{release}
Provides:  python3-cryptography%{?_isa} = %{version}-%{release}
%endif
```

so that installing `python312-cryptography` cleanly obsoletes/replaces the BaseOS-provided `python3-cryptography` on RHEL/CentOS Stream 10 specifically, instead of hard-conflicting with it, while leaving RHEL 9 behavior untouched.

## Also at risk

`python-requests` in this repo has the identical shape — hardcoded `%global python3_pkgversion 3.12`, builds as `python312-requests`, and `foremanctl`'s `pre_install` role also requests it as plain `python3-requests`. No inline `Obsoletes`/`Provides` for the unversioned name exists for it either. Worth adding the same RHEL10-gated pattern to `python-requests.spec` in the same pass, even though we haven't hit a live conflict for it yet.

We checked the rest of that `pre_install` package list (`python3-libsemanage`, `python3-psycopg2`, `python3-requests-oauthlib`) against both `pulpcore-packaging` and `foreman-packaging` — none are independently built there under a versioned name, so they're not at risk of this specific collision (`requests-oauthlib` was previously built under `python3.11`/`python3.12` naming but has since been retired and is already covered by existing `Obsoletes` entries in `pulpcore-obsolete-packages`).

Tracked internally at SAT-49499.

Odilhao:
## Problem

Installing Satellite container images on RHEL 10 fails during `satellitectl pull-images`'s `pre_install` role with an RPM transaction conflict:

```
Unknown Error occurred: Transaction test error:
  file /usr/lib64/python3.12/site-packages/cryptography/... from install of
  python3-cryptography-43.0.0-4.el10.x86_64 conflicts with file from package
  python3.12-cryptography-46.0.7-2.el10pc.x86_64
```

`foremanctl`'s `pre_install` role requests `python3-cryptography` (and `python3-requests`) by the plain, unversioned name. On RHEL 10, `python3` is `python3.12`, and this repo builds both packages under the versioned `python312-*` name via a hardcoded `%global python3_pkgversion 3.12`, so both names resolve to the same physical site-packages path. When a host has both installed, dnf reports a hard file conflict rather than resolving versions across two differently-named packages.

## Fix

Add the same inline `Obsoletes`/`Provides` pattern already used by `python-ldap.spec` for `python3-pyldap`, gated to `%if 0%{?rhel} >= 10` so RHEL 9 (where `python3-cryptography`/`python3-requests` are distinct, legitimate python3.9 packages) is untouched.

Obsoletes are versioned against `%{version}-%{release}`, matching this repo's convention.

`python-requests` has the identical shape (hardcoded `python3.12` build, requested as plain `python3-requests` by the same `pre_install` role) — no live conflict has been hit for it yet, but the collision risk is identical, so it's fixed in the same pass.

Also added `python3-cryptography` and `python3-requests` to `comps/comps-pulpcore-el10.xml` so the published repo group metadata includes the unversioned names these packages now provide/obsolete on RHEL 10 (matching the existing `python3-createrepo_c`/`python3-libcomps` entries already there).

Other packages in that `pre_install` list (`python3-libsemanage`, `python3-psycopg2`, `python3-requests-oauthlib`) aren't independently built under a versioned name here or in `foreman-packaging`, so they're not at risk of this specific collision.

### DIFF
--- a/comps/comps-pulpcore-el10.xml
+++ b/comps/comps-pulpcore-el10.xml
@@ -25,6 +25,7 @@
       <packagereq type="default">pulpcore-selinux</packagereq>
       <packagereq type="default">pyproject-rpm-macros</packagereq>
       <packagereq type="default">python3-createrepo_c</packagereq>
+      <packagereq type="default">python3-cryptography</packagereq>
       <packagereq type="default">python3-libcomps</packagereq>
       <packagereq type="default">python3.12-SecretStorage</packagereq>
       <packagereq type="default">python3.12-aiodns</packagereq>
--- a/comps/comps-pulpcore-el10.xml
+++ b/comps/comps-pulpcore-el10.xml
@@ -27,6 +27,7 @@
       <packagereq type="default">python3-createrepo_c</packagereq>
       <packagereq type="default">python3-cryptography</packagereq>
       <packagereq type="default">python3-libcomps</packagereq>
+      <packagereq type="default">python3-requests</packagereq>
       <packagereq type="default">python3.12-SecretStorage</packagereq>
       <packagereq type="default">python3.12-aiodns</packagereq>
       <packagereq type="default">python3.12-aiofiles</packagereq>
--- a/packages/python-cryptography/python-cryptography.spec
+++ b/packages/python-cryptography/python-cryptography.spec
@@ -9,7 +9,7 @@
 
 Name:           python%{python3_pkgversion}-%{pypi_name}
 Version:        46.0.7
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        cryptography is a package which provides cryptographic recipes and primitives to Python developers
 
 License:        BSD or Apache License, Version 2.0
@@ -34,6 +34,12 @@ BuildRequires:  openssl-devel
 BuildRequires:  gcc
 
 Requires:       python%{python3_pkgversion}-cffi >= 2.0.0
+
+%if 0%{?rhel} >= 10
+Obsoletes: python3-%{pypi_name} < %{version}-%{release}
+Provides:  python3-%{pypi_name} = %{version}-%{release}
+Provides:  python3-%{pypi_name}%{?_isa} = %{version}-%{release}
+%endif
 
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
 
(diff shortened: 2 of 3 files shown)


Odilhao:
[test rpm]

ogajduse (approved):

Odilhao:
[test rpm]

